PR #370, opened 2021-04-14
LH: add noopener to external links

estelle:


mikenikles (approved):
Thank you Estelle! LH score++ 😄

PR #389, opened 2021-04-21
Title changes based on documentation page

asharonbaltazar:
Let me just preface by saying Svelte makes my head spin. 

Objective: 
Have the document title change based on the documentation title.

Possible solution: 
Add a `title` property to each `.md` file and import it inside `docs-content-layout`, and pass that as a prop to <OpenGraph />. I didn't go with this because it meant I would have to update each `.md` file to include the property `title`. Maybe there's a better way...

Solution:
Have an `onMount` that grabs the `innerText` of the first `h1`, which is present on every `.md` file, and pass that as a prop to the `<OpenGraph />` component. 

Blockers: This implementation isn't done yet. Between renders, it's possible to see the fallback text flash for a second before the correct title appears. I'm unfamiliar with Svelte, so therefore I'm requesting a little bit of help. 
@jankeromnes  @gtsiolis

### DIFF
--- a/src/components/docs/docs-content-layout.svelte
+++ b/src/components/docs/docs-content-layout.svelte
@@ -1,7 +1,9 @@
 <script>
+  import OpenGraph from "../open-graph.svelte";
   import docsCurrentSectionStore from "../../stores/docs-current-section";
 
   export let section;
+  export let title;
 
   $: $docsCurrentSectionStore = section;
 </script>
@@ -10,6 +12,14 @@
   @import "../../assets/docs";
 </style> -->
 
+<OpenGraph
+  data={{
+    description:
+      "Explore the documentation to learn more about Gitpod.io and Gitpod Self-Hosted.",
+    title: title ? title : "Gitpod Documentation",
+    type: "website",
+  }}
+/>
 <div class="content-docs">
   <slot />
 </div>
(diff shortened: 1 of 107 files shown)


asharonbaltazar:
Opted instead to add title to each `.md` file since this is the method used in regards to the blog `.md` files and I wanted to keep things consistent and clean without any hacky methods. It works beautifully too

ghuntley:
Thanks for stepping up to help with this @asharonbaltazar. Have requested a review by someone who is familiar with Svelte.

src/components/docs/docs-content-layout.svelte
```diff
@@ -10,6 +12,14 @@
   @import "../../assets/docs";
 </style> -->
 
+<OpenGraph
+  data={{
+    description:
+      "Explore the documentation to learn more about Gitpod.io and Gitpod Self-Hosted.",
+    title: title ? title : "Gitpod Documentation",
+    type: "website",
```
mikenikles, line 20: Looking at that I notice there is no need for `type` to be a mandatory prop. Tracking this as a separate issue at #399.

mikenikles (approved):
Excellent, thank you so much for the contribution 🎉 !

asharonbaltazar:
Of course! I'm always down for anything simple.

Issue #377, opened 2021-04-16
500 error when accessing https://www.gitpod.io/docs


shaal:
### Bug description

The URL https://www.gitpod.io/docs (or any other direct link to documentation) cannot be accessed directly and results in a 500 error.
The workaround is going to any other page (like the homepage), and then click on 'Docs' on the menu.

### Steps to reproduce

Open https://www.gitpod.io/docs in your browser, you'll get a 500 error.
![image](https://user-images.githubusercontent.com/22901/114972137-e6fce000-9e4b-11eb-8821-bbb63b103929.png)

### Expected behavior

I expect to be able to open directly any page in documentation.

nisarhassan12:
Thanks, @shaal for reporting this. I'm unable to reproduce this. 
![Peek 2021-04-16 11-29](https://user-images.githubusercontent.com/46004116/114981205-0ffe8e80-9ea7-11eb-9b1c-f1a4d5918718.gif)

Can you please share screenshots of the Network tab and the browser console after you hard reload the docs page?

AlexTugarev:
@nisarhassan12, I've seen the same error yesterday. 

Could it be, that during a deployment/update of the website the service is unavailable?

mikenikles:
I can reproduce it, although not reliably. It happened regularly with Firefox on MacOS, but a few minutes later it worked. Same for Brave on Android, broken for 10 or so reloads, then paused for a few minutes and retried successfully.

I suspected a new deployment as well at first, but a full page reload should work regardless of whether a deployment happened recently. I'll dig a bit deeper to see what's going on.

mikenikles:
Hey @shaal, could you clear your local cache and give www.gitpod.io/docs another go please?

I ended up upgrading our application framework (PR #379) as there were a few recent bug fixes that allowed me to switch to static pages rather than server-side rendered. I _suspect_ we had an issue with SSR in a previous deployment.

ghuntley:
Gitpod needs to decide what to do about `www.gitpod.io` (the website) and `gitpod.io` (the app). Apologies @shaal, know that I also share this frustration and run into it when sharing links on social media. Google Chrome flattens `www.gitpod.io` into `gitpod.io` which also exacerbates the problem. 

The upstream problem is https://github.com/gitpod-io/gitpod/issues/3898.

shaal:
Somewhat related? going to https://gitpod.io/docs (without 'www' prefix) results in 404 error.

Same with -
https://gitpod.io/features
[...]

PR #394, opened 2021-04-21
Pre-render all pages

mikenikles:
Relates to #377 

Tasks:
- [x] Remove all trailing slashes on internal links
- [x] Enable prerender for the remaining pages in /docs and /blog

In case anyone else needs to insert the [SvelteKit prerender instructions](https://kit.svelte.dev/docs#ssr-and-javascript-prerender) to a bunch of files, here's a script:

```sh
#!/bin/sh
for md in $(find . -name '*.md'); do
  { echo '<script context="module">\n  export const prerender = true;\n</script>\n'; cat "$md"; } >"$md.tmp" && mv "$md.tmp" "$md"
done
```

If you do this to Markdown files that have frontmatter, you have to also reorder the `<script>` and the frontmatter. See [this tweet](https://twitter.com/mikenikles/status/1385019268052422666/photo/1).


src/components/docs/mobile-menu/sub-menu.svelte
```diff
@@ -6,7 +6,7 @@
   import topicsState from "../states/topics-state";
   import subMenuState from "../states/sub-menu-state";
 
-  export let currentSection;
+  export let currentSection: any = {};
```
mikenikles, line 9: The first set of release notes are not displayed in the docs navigation and don't have a corresponding section. This `{}` default value deals with that case.

Issue #404, opened 2021-04-25
Close button overlaps content on smaller screens


asharonbaltazar:
I'm unsure if there's a duplicate issue, despite my quick glance at the issues, so if it is, just delete this 😋.
While I worked on PR #403, I noticed the close button (X) in the top banner overlapped content on smaller screens. 

![image](https://user-images.githubusercontent.com/58940073/115979876-1a2b2780-a54e-11eb-90d4-e6278a813325.png)

If this issue isn't a duplicate, I'd like to be assigned it. Thank you.

nisarhassan12:
Thanks @asharonbaltazar for reporting this please feel free to make a PR. 

For testing you can comment out line#7 and line#11 in https://github.com/gitpod-io/website/blob/main/src/routes/%24layout.svelte

nisarhassan12:
Fixed in #443

PR #443, opened 2021-05-09
Overlapping button solution

asharonbaltazar:
Addresses issue #404 

The current layout and styling of the banner made it difficult to find a satisfactory solution. I opted to separate the megaphone emoji and placed it in its own separate `<span>` element. I also changed the `absolute` positioning of the close button to `flex-1`. In fact, the `flex` property is now present on every element in the Announcement banner to help distribute them evenly. This is the result on mobile screen sizes:
image
This is the result on a desktop:
image

Don't like it? I'm open to suggestions. 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/443"></a>

### DIFF
--- a/src/components/banners/announcement.svelte
+++ b/src/components/banners/announcement.svelte
@@ -11,12 +11,13 @@
 </script>
 
 <Banner storageKey="announcement-{announcementDate}" let:closeBanner>
-  <p class="relative flex-1 text-center">
-    📣 DevX Conf - Bring back joy and speed to our workflows. | <a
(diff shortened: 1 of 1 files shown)


ChristinFrohne:
Thanks a lot for your contribution @asharonbaltazar! Currently, we don't have a banner on the website anymore, as the event has already passed. But we should keep your changes for the next time we'll enable the banner. 
Can you take care of that @nisarhassan12 ?

nisarhassan12 (approved):
Thanks @asharonbaltazar for the PR. LGTM.

PR #463, opened 2021-05-20
preparing for 0.9.0 release

solomonope:
<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/463"></a>


mikenikles:
Hey @solomonope, we have to add a `src/routes/docs/self-hosted/latest` directory with the 0.9.0 docs. When the next version gets released, we will rename `latest` to `0.9.0`.

Do you already have 0.9.0 docs anywhere? Otherwise, we can start with copying `0.8.0` to a new `latest` directory.

solomonope:
> Hey @solomonope, we have to add a `src/routes/docs/self-hosted/latest` directory with the 0.9.0 docs. When the next version gets released, we will rename `latest` to `0.9.0`.
> 
> Do you already have 0.9.0 docs anywhere? Otherwise, we can start with copying `0.8.0` to a new `latest` directory.

Hey @mikenikles  I have added `0.9.0` docs to the latest folder

mikenikles:
Thanks for that, I noticed a few things:
1. `install/oauth.md` no longer exists in the `latest` directory, is that on purpose? If not, we have to revert where I [removed it from the navigation](https://github.com/gitpod-io/website/pull/463/commits/9b2446eec5f5efda5b923b51d4f1bc6fcdfbb3fa#diff-cc35b177d1bf6e55ea29dfe9218dafd38abdd2fd5d82bc312e6ef3495158909dL97).
2. A new `install/install-on-gcp.md` exists which replaces the existing `install/install-on-gcp-script.md` file. Is this correct? If so, we'll have to update the navigation in `src/routes/docs/$layout.svelte`.
3. The `install/install-on-gcp.md` links to a relative path `../../../install/gcp-terraform`. This won't work on the website.

The 3. item above made me wonder: For self-hosted installations, do we include the docs from the `gitpod-io/gitpod` repo? If that's the case, we need to have a chat because I want to have a single source of truth and at the moment, the latest docs are in the `gitpod-io/website` repo.

solomonope:
@geropl  Please chime in on the coversation on the `install/oauth.md`  I reckon it is no longer required since we have a dialog that we use to  set this up, immediately after installation.

src/routes/docs/self-hosted/latest/install/upgrade.md
```diff
@@ -9,13 +9,33 @@ title: Gitpod Self-Hosted Upgrade Notes
...
+  auth:
+    username: your-rabbitmq-user
+    password: your-secret-rabbitmq-password
+```
+
+If this is not present, `helm` will fail with the following message:
+
+> rabbitmq username is required, please add a value to your values.yaml or with the helm flag --set rabbitmq.auth.username=xxxxx
+
+## Upgrading Gitpod from v0.6.0 to v0.8.0
```
geropl, line 32: 0.6.0 to 0.7.0

Also, the refenreces below should stay at 0.7.0

geropl:
> Please chime in on the conversation on the install/oauth.md I reckon it is no longer required since we have a dialog that we use to set this up, immediately after installation.

@solomonope Ah, yes. Makes sense to delete it now that we have the setup again :+1:

geropl (approved):
Thanks for working this out!

I have two comments but already approve to unblock here. Please have a look at those before merging.

solomonope:
> Thanks for working this out!
> 
> I have two comments but already approve to unblock here. Please have a look at those before merging.

@geropl all the feedback has been resolved

PR #538, opened 2021-06-02
Initial version of local app blog post

rl-gitpod:
TODO: replace reused placeholder teaser image

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/538"></a>

[Rendered](https://github.com/gitpod-io/website/blob/rl/blog-post-for-local-app/src/routes/blog/local-app.md)


svenefftinge:
Cool! It would be good to provide real links (at least for the download). Also, some (bullet point list, source code) of the formatting doesn't look too nice. I wonder if that is because of wrong markup or the general styling in our blog? cc @mikenikles

src/routes/blog/local-app.md
```diff
@@ -0,0 +1,98 @@
...
+title: Accessing Gitpod workspaces locally
+---
+
+<script context="module">
+  export const prerender = true;
+</script>
+
+### Connecting to your workspaces
+
+While Gitpod can seamlessly integrate into your workflow in the vast majority of cases, there are times where you may want to access a workspace from outside its confines. While this has been possible using the existing VS Code port forwarding feature, it only worked for an HTPP-based protocol.
```
akosyakov, line 18: We could not use VS Code Remote in the past and cannot use right now. I would rather talk about limitations with having only 2 ports in the past. And it would be better to just allow to have the same localhost but remotely.

src/routes/blog/local-app.md
```diff
@@ -0,0 +1,98 @@
...
+  GITPOD_HOST=https://gitpod.io ./gitpod-local-companion-[darwin|linux|windows] --mock-keyring
+
+
+```
+
+If you are not logged in to Gitpod, it will take you through that flow. If you haven't run the local app previously it will also ask you to approve access to the control data of all your workspaces, returning a token that grants the access. This token will be stored in your local keyring for future use (unless you specify otherwise).
+
+Once approval has been granted it will connect to your currently active workspaces and allow VSCode to control the tunnelling between the remote workspace and your local environment using the standard VSCode Ports View.
+
+#### Extended VSCode Ports View
```
akosyakov, line 74: we should not mention VS Code ports view, explain how to use our instead

src/routes/blog/local-app.md
```diff
@@ -0,0 +1,98 @@
...
+  export const prerender = true;
+</script>
+
+### Connecting to your workspaces
+
+While Gitpod can seamlessly integrate into your workflow in the vast majority of cases, there are times where you may want to access a workspace from outside its confines. While this has been possible using the existing VS Code port forwarding feature, it only worked for an HTPP-based protocol.
+
+Gitpod is pleased to announce a preview release of the _Gitpod Local Companion_ app that is designed to allow localhost access to any TCP port in a remote workspace regardless of protocol.
+
+The app is intended to run locally and enables automatically tunneled connections to your workspace, either privately or, if required, publicly (on your laptops IP for example).
```
akosyakov, line 22: It would be good to mention somewhere that by default we are trying to tunnel on the same port, but otherwise we pick random. A user can see it in our ports view. Probably should go under `Extended Ports View`

src/routes/blog/local-app.md
```diff
@@ -0,0 +1,98 @@
...
+
+This opens up any environment or framework that assumes localhost access. Also linking to http://localhost:5000 in your projects README.md will now _just work_ for anyone clicking on the _Gitpod_ button on your project.
+
+As a preview release, not all of the features are implemented, most notably the tunneling is 1-way only: local -> workspace.
+Of course, we have big plans! Future features may include:
+
+- Listing what workspaces are running
+- Listing what ports are currently tunneled and in which direction (workspace to/from local)
```
akosyakov, line 30: reverse tunneling is not supported yet

src/routes/blog/local-app.md
```diff
@@ -0,0 +1,98 @@
+---
+author: rl-gitpod
+date: Mon, 31 May 2021 18:00:00 UTC
+excerpt: Accessing Gitpod workspaces locally
```
akosyakov, line 4: I feel somehow that title should contain localhost. First someone can think that locally is about offline too, and second that you have the same setup on localhost as with local development removes a friction dealing with `gp url` and so on to get local setups to run in Gitpod.

akosyakov:
I think it would be cool to have some recording and explanation around a project like https://github.com/akosyakov/parcel-demo. It shows also limitations which have now and we can make a story how local app resolves it.

rl-gitpod:
> Cool! It would be good to provide real links (at least for the download). 

What dev would prefer a link over using curl ;-)

Jokes aside, good point, I will replace.

> Also, some (bullet point list, source code) of the formatting doesn't look too nice. I wonder if that is because of wrong markup or the general styling in our blog? cc @mikenikles

Agreed.
I think it is general e.g. these bullet points from the [latest entry](https://www.gitpod.io/blog/journey-to-site-reliability-engineering-at-gitpod) 

Screen Shot 2021-06-03 at 10 42 09 am

src/routes/blog/local-app.md
rl-gitpod: I was referring to making the port public and using the resultant URL which I thought was already available but I realise now that I was conflating code spaces/vs code remote/gitpod (hence the next section).

> I would rather talk about limitations with having only 2 ports in the past.  And it would be better to just allow to have the same localhost but remotely

Only 2? 'same localhost' - can you elaborate?

src/routes/blog/local-app.md
rl-gitpod: Understood - see above comment

src/routes/blog/local-app.md
rl-gitpod: Right.
It's under the "Future features may include". I'll highlight the *may* a little more.

rl-gitpod:
> I think it would be cool to have some recording and explanation

I agree, but I wanted to get it out for comment before I slept.

I've gone back and forth on a recording/example, (with myself and Geoff), trying to find something really compelling for local app (in its current state) without much success. After getting to the end of the first draft I became a little concerned that it was getting too long. On that front, looking at the latest [SRE entry](https://www.gitpod.io/blog/journey-to-site-reliability-engineering-at-gitpod) we're ok.

> around a project like https://github.com/akosyakov/parcel-demo. It shows also limitations which have now and we can make a story how local app resolves it.

My issue is I'm not 100% sure of what the limitations existed prior to local app. I get that if you look at the terminal output

`Server running at http://localhost:1234` 

and simply copy the url from the output into your browser it will fail. However, there is the mouse over as below:

Screen Shot 2021-06-03 at 11 22 57 am

I'm assuming that existed prior to local app and worked around that issue to an extent?

On the compelling example side, the other issue is that we now have docker working in workspaces (at least it wfm in limited trials) which takes out a complete class of problems i.e. why connect to a local docker when you can run it in the workspace? 

The less ubiquitous/compelling use case of home automation remains which can be tricky to setup. Hrm, I guess I could do an example connecting to my off-grid monitoring but how applicable is that across the board?

Definitely possible (very likely??) I'm overthinking this and the simple parcel example is sufficient - happy to hear suggestions!

src/routes/blog/local-app.md
rl-gitpod: I see your point wrt offline. Maybe:

/locally/as localhost/ 

> a friction dealing with gp url

As per the compelling use case comment I just added, I wasn't familiar with what this friction was, but looking at the (now ancient) [Gitpodify-ing guide)](https://www.gitpod.io/blog/gitpodify/) I understand it better. 
I'll take a pass at highlighting this, perhaps with a reference to that guide.

src/routes/blog/local-app.md
akosyakov: As far as I know Gitpod can serve only on 2 real ports: 80 and 443. We remap other ports as a suffix to workspace host to create an external URL, i.e. 9000-my-workspace:443. As a result some projects should be reconfigured to use external URLs (via gp url). Sometimes it is not possible since underlying frameworks are not configurable enough like parcel. See for instance https://github.com/gitpod-io/gitpod/issues/1881 and https://github.com/gitpod-io/gitpod/issues/3282. With the local companion app such reconfiguration is not necessary anymore. It works the same way as with the local setup.

akosyakov:
> My issue is I'm not 100% sure of what the limitations existed prior to local app. I get that if you look at the terminal output
> Server running at http://localhost:1234
> and simply copy the url from the output into your browser it will fail. However, there is the mouse over as below:

If you click on it holding cmd then it will automatically translate such URL in external URL, i.e. 1234-my-workspace:443 in case if there is no local tunnelling, or localhost:{remotePort} otherwise, probably 1234 

> Definitely possible (very likely??) I'm overthinking this and the simple parcel example is sufficient - happy to hear suggestions!

Yeah, I meant for recording it could be something simple to show the flow. It is better also don't use my repo, but setup some parcel example under http://github.com/gitpod-io then and add a proper README referencing this post.

src/routes/blog/local-app.md
akosyakov: some ideas:
`Gitpod Local Companion: Localhost is going remote`
`Unleashing localhost with Gitpod Local Companion`

We should really think about naming, it is way too long :)

rl-gitpod:
We don't appear to have used Loom in the blog prior to this. Do we want to move it to youtube?

JohannesLandgraf:
Nice!! Some high level feedback:

- It feels strange to start a blog post with a title, there should be some introduction similar to our other blog posts. 
- We deviate quite a bit from our branding and styling wrt to hero image, quality of screenshots etc. Looping in @ChristinFrohne here to give some guidance wrt to styling and branding. 
- To highlight code snippets even more you could use the styling that we use e.g. in https://www.gitpod.io/docs/getting-started. 
- Would be great to have an Outlook / What is next section where talk about possibilities this opens for us and where we want to take this
- The screencast should be on Youtube and should be recorded with highest res possible

ChristinFrohne:
some additional feedback regarding the layout:
- It'd be great if you use the standard headline structure. Currently, there is no `<h2>`. 
- the bullet points are not aligned with the text. Also, it'd be best to use em dashes `—` instead of `-`. 
- for the header image, [here](https://www.figma.com/file/Cbn7RdOyAsUkNBpGFgQjxL/Brand-assets?node-id=5%3A238) are some guidelines. E.g is there an existing 3D illustration that you could use?
- any reason why you changed to a smaller font size here? 👇  If not, we should to stick to the default style
Bildschirmfoto 2021-06-03 um 15 49 31

src/routes/blog/local-app.md
rl-gitpod: Ah, ic what you meant by 2 ports - thanks.

rl-gitpod:
> Nice!! Some high level feedback:

Thanks.
> 
> * It feels strange to start a blog post with a title, there should be some introduction similar to our other blog posts.

If I was pedantic I would say all the blog posts have titles ;-) It is just that we have titles specified in a different way to 'standard' markdown (I know that is an oxymoron). 

tbh I've been focused on content to date. I'm glad I can now focus on layout - rectified.

> * We deviate quite a bit from our branding and styling wrt to hero image, quality of screenshots etc. Looping in @ChristinFrohne here to give some guidance wrt to styling and branding.

Yes, hence the > TODO: replace reused placeholder teaser image. I look forward to the documentation - it will help everyone going forward.

> * To highlight code snippets even more you could use the styling that we use e.g. in https://www.gitpod.io/docs/getting-started.

If you mean the yaml list then it is using the same (even specifying language yaml). Not sure why it isn't rendering locally in the same way. Or do you mean the 'brand-light' background as below?

Screen Shot 2021-06-04 at 3 17 52 pm

> * Would be great to have an Outlook / What is next section where talk about possibilities this opens for us and where we want to take this

There was a 'Future plans include' paragraph but I agree making it a section will highlight that further - done, thanks. @svenefftinge / @csweichel  - I'd appreciate some input on additional future plans if possible?

> * The screencast should be on Youtube and should be recorded with highest res possible

Is the process documented in Notion (I couldn't find it with a quick search)? What about access to the Gitpod YT account?

rl-gitpod:
> some additional feedback regarding the layout:
> 
> * It'd be great if you use the standard headline structure. Currently, there is no `<h2>`.

One of the gotchas' with markdown there is no 'h2' per se. I have a habit of avoiding using '## ' as it is often rendered too large on laptop screens - all are now '## '

> * the bullet points are not aligned with the text. Also, it'd be best to use em dashes `—` instead of `-`.

Interesting. That was copied from other blog posts. I see there is an issue #561 around it.

> * for the header image, [here](https://www.figma.com/file/Cbn7RdOyAsUkNBpGFgQjxL/Brand-assets?node-id=5%3A238) are some guidelines. 

I don't have access to that page with my Gitpod.io login but I can see the 'Projects & Teams' assets 🤷 

> E.g is there an existing 3D illustration that you could use?

From where e.g. do we have an account on one of the asset sites?

> * any reason why you changed to a smaller font size here? 👇  If not, we should to stick to the default style

I *actually* wanted a subscript as it totally dominates the list. I agree the quick hack I tried looks terrible - reverted to default

src/routes/blog/local-app.md
rl-gitpod: @akosyakov is the random port selection working for you or am I misunderstanding what you're saying?

If I run a local python server on 127.0.0.1:3000 I can see that the local app tunnels port 3000 from the website workspace on a random port as expected:

> INFO[0003] tunnel[host:3000]: listening on 127.0.0.1:59820...  workspace=pink-damselfly-mv5i7w13

But regardless of public/private toggle in the Ports view, opening the [browser link in the ports view](https://59820-pink-damselfly-mv5i7w13.ws-us08.gitpod.io/) for port 3000 ends up with:

> Port 59820 Not Found

It only works if I go direct to localhost:59820 *unless* the 'Tunnel on all interfaces' is on in which case localhost:3000 works as expected but not the browser icon.

Or do we need to highlight this fact in the blog entry i.e. you need to tunnel on all interfaces'?

rl-gitpod: Also wrt "A user can see it on our ports view"...

It isn't obvious unless you know what to look for in the local terminal:

Screen Shot 2021-06-07 at 4 00 25 pm

or use the Debug view:

Screen Shot 2021-06-07 at 4 00 34 pm

akosyakov: I meant that if nothing is running on port 3000, then we will tunnel on 3000 locally. But if it is already reserved we will use random port then it should be displayed in the ports view as something like `3000:62572`.

akosyakov: It was fixed, please download the latest app from gitpod.io and try again.

src/routes/blog/local-app.md
```diff
@@ -0,0 +1,124 @@
...
+
+```shell
+  ./gitpod-local-companion-[darwin|linux|windows] --mock-keyring
+
+
+```
+
+If you are not logged in to Gitpod, it will take you through that flow. If you haven't run the local app previously it will also ask you to approve access to the control data of all your workspaces, returning a token that grants the access. This token will be stored in your local keyring for future use (unless you specify otherwise).
+
+Once approval has been granted it will connect to your currently active workspaces and allow VSCode to control the tunnelling between the remote workspace and your local environment using the standard VSCode Ports View.
```
akosyakov, line 83: Please use `Ports View` instead of `standard VSCode Ports View`. Users who is aware of vscode-remote will be confused by it. Standard VS Code ports view is somethign which we are trying to hide. The same in `Extended Ports View` section.

src/routes/blog/local-app.md
rl-gitpod: I've added this to the blog entry with image

src/routes/blog/local-app.md
rl-gitpod: I've switched to using Remote Explorer Ports View since that is what is used by VSCode and is quite visible in the screenshots :-)

svenefftinge:
> @svenefftinge / @csweichel - I'd appreciate some input on additional future plans if possible?

Sorry, I missed this ping, but what you wrote sounds good. I have stopped using github notifications some time ago as there are just too many 🙄 .

svenefftinge (approved):
🚀

PR #556, opened 2021-06-03
[docs]: improve the design of the feedback widget.

### LINKED ISSUE #555: [docs]: improve the design of the feedback widget
image

I'll make a PR as per discussed internally with Christin in our 1:1 meeting.

nisarhassan12:
Fixes #555

This is how it looks now:

| Before | After |
| - | - |
| image | image 
| image | image |

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/556"></a>

### DIFF
--- a/src/components/docs/feedback-widget.svelte
+++ b/src/components/docs/feedback-widget.svelte
@@ -35,55 +35,62 @@
   }
 </style>
 
-<div>
-  <h5 class="mt-12 mb-6 text-center">Was this helpful?</h5>
-  {#if resultMessage}
-    <p class="text-center">{resultMessage}</p>
-  {:else}
-    <form
-      on:submit|preventDefault={submitFeedback}
-      title="Share your feedback about our docs"
-    >
-      <div class="flex justify-center space-x-6">
-        {#each new Array(4) as _, index}
-          <button
-            on:click|preventDefault={() => (selectedEmotion = index + 1)}
-            class:selected={selectedEmotion === index + 1}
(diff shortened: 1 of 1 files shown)


src/components/docs/feedback-widget.svelte
```diff
@@ -35,55 +35,62 @@
...
-              autocorrect="off"
-              type="text"
-            />
-          </div>
-          <div>
-            <span
-              ><button role="button" type="submit" class="btn-primary"
-                ><span>Send</span></button
-              ></span
+<div class="my-huge">
```
mikenikles, line 38: I don't see this class used anywhere, can we remove it?

mikenikles (approved):
Very well done, this looks a lot cleaner. Thank you Nisar!

src/components/docs/feedback-widget.svelte
nisarhassan12: Thanks. I added it for 2 reasons one I wanted spacing huge to be applied across the y-direction I could have added it to the white box itself but the main usage of this is being able to use `mx-auto` on the white box to centre it.

mikenikles: Oh my 🤣... I thought this was a custom CSS class name called `my-huge` as in some kind of test name. Reading your comment, I now realize this is "margin", "y" with a custom size "huge".

If I ever need a good reason to showcase why using Tailwind's default, digit-based scale is better than a name-based one, this right here is it 😄.

PR #624, opened 2021-06-14
fixes 544 - left navbar selected entry inconsistency

opensas:
The problem was that the original regex `/\/docs\/$/.test(menuItem.path)` (see [here](https://github.com/gitpod-io/website/blob/main/src/components/docs/menu-section.svelte#L13)) required a trailing slash to match, so `https://www.gitpod.io/docs` didn't match and `isActiveSection` was false for the Introduction docs page. Making the trailing slash optional with `/\/docs\/?$/.test(menuItem.path)` solved the issue:

image

To implement the b) option (never use white bubble if it's a menu item with no subitems) I added the following condition:

```svelte
  <div
    class:isActiveSection={isActiveSection && menuItem.subMenu}
    class="menu-container"
  >
```

Which yields this result:

image

Personally I prefer the a) option, I find it more clear, consistent and better looking. To go that way just leave that line like it currently is (see [here](https://github.com/gitpod-io/website/blob/main/src/components/docs/menu-section.svelte#L44))

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/624"></a>

### DIFF
--- a/src/components/docs/menu-section.svelte
+++ b/src/components/docs/menu-section.svelte
@@ -10,7 +10,7 @@
           ? $docsCurrentSectionStore.replace(/\d\.\d\.\d/, "latest")
           : $docsCurrentSectionStore
       ) >= 0
-    : /\/docs\/$/.test(menuItem.path);
+    : /\/docs\/?$/.test(menuItem.path);
 </script>
 
 <style lang="scss">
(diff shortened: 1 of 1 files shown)


JohannesLandgraf:
LGTM and I'd also vote for option a) which is consistent

nisarhassan12:
Thanks, @opensas for the Pr it LGTM once it is merged I will take care of implementing `a` as you and Johannes suggested.

mikenikles:
@nisarhassan12 could you take care of option a) please as discussed in this PR 🙏

PR #627, opened 2021-06-14
fixes #192 - autoclose faq

### LINKED ISSUE #192: Automatically close the opened FAQ when another one is opened.
To save space we should close the opened FAQ when another one is opened.

opensas:
This one ended up being more tricky than expected

I has to use a FaqContainer to set a store and a context

I'm not very convinced about the FaqsContainer name, I would just use Faqs for the container and Faq for each question, but I didn't want to create confusion with pricing/faqs and self-hosted/faqs.

More over, I would move FaqsContainer and Faq to their own faqs folder, now that it is being used in two different places

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/627"></a>


mikenikles:
Thanks for looking into this! I got inspired to see if there's a way to drop the reactive `isActive` variable. What do you think of this solution https://svelte.dev/repl/289109bf302943a29c657d4bfe31924f?version=3.38.3?

> I'm not very convinced about the FaqsContainer name, I would just use Faqs for the container and Faq for each question, but I didn't want to create confusion with pricing/faqs and self-hosted/faqs.

👍 At this point, it doesn't even have to include "Faq" I would think and it could just be an `Accordions` and `Accordion` component. A generic component we can use to create disclosure elements with arbitrary content. Thoughts?

> I would move FaqsContainer and Faq to their own faqs folder

💯 Totally agreed, let's do that.

opensas:
> Thanks for looking into this! I got inspired to see if there's a way to drop the reactive `isActive` variable. What do you think of this solution https://svelte.dev/repl/289109bf302943a29c657d4bfe31924f?version=3.38.3?

yes, that was also my first attemp, but then I found out that if I click on the open faq the component would end up in an inconsistent state:

image

that is, no faq open, but $activeFaq pointing to the last opened faq

The problem is that if you just replace on:toggle with `on:toggle={({target}) => {$activeFaq = target.open ? summary : null}}` when you click on a different faq it won't work, because on:toggle gets called first for the faq that is being closed, which sets $activeFaq to null, leaving no faq opened.

So I had to do something like [this](https://svelte.dev/repl/ca31c007e8ee49f9856b4a1cbcb2770b?version=3.38.3):

```
<details open={$activeFaq === summary} on:toggle={({target}) => {
	$activeFaq = target.open ? summary : ( $activeFaq === summary ? null : $activeFaq )
}}>
```

and that's when I though it would be more clear to extract that to a setActive function and also create the isActive var, I found it more readable.

Tell me what you think about it. 

OTH I like tha accordion/s approach, I'll create a PR for that

mikenikles:
Thanks for the additional explanation. Let's keep it the way you developed it for this PR and if we get to refactoring the components into accordions, we'll do that as a follow up PR.

Issue #648, opened 2021-06-19
Fully migrate the codebase to use Tailwind. 


nisarhassan12:
We should get rid of all the custom CSS to get rid of these kinds of performance issues https://github.com/gitpod-io/website/issues/643#issuecomment-863764621

- [x] avatars.scss => #1481
- [x] base.scss => #1488
- [x] blog_preview.scss => #1489
- [x] blog.scss => #1490
- [x] buttons.scss
- [x] cards.scss => #1500 
- [x] error_page.scss => #1486
- [x] feature_table.scss => #1516 
- [ ] forms.scss
- [x] images-across.css => #1487
- [x] layout_helpers.scss => https://github.com/gitpod-io/website/pull/1520
- [x] modal.scss https://github.com/gitpod-io/website/pull/1493
- [x] reset.scss => #1470 
- [x] screencasts.scss => #1491
- [x] team.scss => #1485
- [x] text_blobs.scss => #1492 
- [ ] typography.scss => Awaiting https://github.com/sveltejs/svelte/issues/2324
- [ ] variables.scss
- [x] blog.scss => #1491
- [x] markdown-common.scss => #1490
- [ ] styles.scss

`global.scss` is getting renamed to `gloabl.css` and used for the tailwind-layers

mikenikles:
This is something we need to be very careful that it doesn't get out of control. We have quite a bit of custom CSS, removing pieces may impact parts of the website we not immediately consider.


The work for this issue has to happen in multiple, small PRs as much as this is possible. For example, the following are examples of individual PRs:
* Header
* Footer
* Docs left-side nav
* Docs & blog post content
* Titles across all pages
* Pricing cards
* Left / right content with images and text, such as used on the homepage and features pages

nisarhassan12:
> Left / right content with images and text, such as used on the homepage and features pages

I have addressed this in https://github.com/gitpod-io/website/pull/719/commits/a4a3935ee584aae1d13aa9fba76ec3fea18144e4

mikenikles:
We are going to keep this open for tracking purposes. We migrate individual components and pages to Tailwind when we touch them as part of other issues.

Over time, we will get to a point where we have minimal custom CSS. At that time, we will spend a day or two to migrate the remaining custom CSS and close this issue.

JohannesLandgraf:
How is this progressing? We wanted to do this after we launched the website in March :)

Cahllagerfeld:
[...]

PR #660, opened 2021-06-21
Migrate the header to Tailwind.

mikenikles:
Relates to #648.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/660"></a>


nisarhassan12:
Many Thanks, @mikenikles for taking care of this. There are a few things that still need to be addressed:

- [ ] The header no longer has a max-width of `1500px` like it is the case on the production site causing the header to take up `100%` width on all the viewports. It does not look good on my `38inch` wide monitor. Tailwind site also has it constrained to `90rem`

- [ ] The login button initially has a font size of `14px` on desktop as per the specs and `12px` max-width `1140px` when we switch font sizes to a smaller scale for smaller devices in a media query defined in `_variables.scss` as per the [specs](https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=1%3A4490) also the same is true for nav items they have a `14px` font size. 

- [ ] On focus state the colour for the login button turns into complete black + the dimensions for the login button on the mockup and prod are `85 x 32` on your Pr it is `96 x 32`.

mikenikles:
I addressed the comments and aligned as closely as possible with the spec without the need to introduce a `<style>` element with custom styles.

I take full responsibility for the content of this PR 😄.

src/components/main-nav/index.svelte
```diff
@@ -29,28 +29,31 @@
...
     },
   ];
 </script>
 
 <nav
   id="choose-project-observer-target-top"
-  class={`nav text-small ${$menuState ? "bg-off-white" : ""}`}
+  class={`${
+    $menuState ? "bg-off-white " : ""
+  }mx-auto w-full max-w-7xl sm:bg-sand-light`}
```
nisarhassan12, line 42: Thanks. @mikenikles We are getting quite close to getting this change live 🚀 . 

`max-w-7xl` (1280px) is not gonna work for us see the specs for Desktop the canvas has a size of 1440px as per the mad specs and on the live site we currently use `1500px` as the max-width for the nav to look closer to the mockup:

| Before | After |
| - | - |
| image | image |

src/components/main-nav/login-button.svelte
```diff
@@ -1,11 +1,6 @@
...
-    justify-content: center;
-    height: 2rem;
-    width: 5.313rem;
-  }
-</style>
-
-<a href="https://gitpod.io/login/" class="btn-tertiary"> Login </a>
+<a
+  href="https://gitpod.io/login/"
+  class="flex items-center justify-center h-8 w-20 rounded-xl bg-black text-off-white text-sm focus:text-off-white focus:bg-black-hover hover:text-off-white hover:bg-black-hover"
```
nisarhassan12, line 3: We need to use the bolder font-weight here as per the specs i.e `700`.

nisarhassan12:
> I addressed the comments and aligned as closely as possible with the spec without the need to introduce a <style> element with custom styles.

> I take full responsibility for the content of this PR 😄.

Thanks, @mikenikles I'm fine if we merge it like but in my personal opinion is Maybe we should re-introduce the style tag here the menu now looks quite broken on tablet size i.e there is now an overflow happening and the only solution I see here is using a custom breakpoint here then we could also address the overlapping issue that we have on the current production site on tablet size devices.

As a general rule, we could try our best to avoid using style tags and only use them in places where they are absolutely required i.e where not using them would hurt the UX or the appearance. 

Your call here Sorry If I misunderstood you. 

### Before 

at width `775px`

Screen Shot 2021-06-29 at 5 58 27 AM

at width `640px`

Screen Shot 2021-06-29 at 6 08 19 AM


### After

at width `775px`

Screen Shot 2021-06-29 at 5 57 40 AM


at width `640px`

Screen Shot 2021-06-29 at 6 06 40 AM

nisarhassan12:
One small nit the space between the mobile menu items is no longer intact with the mockup on the mockup it is `24px` i.e `var(--xx-small)` and here it is different. It won't be an issue if we leave it like this.

src/components/main-nav/index.svelte
mikenikles: Fixed in 5446efa

src/components/main-nav/login-button.svelte
mikenikles: Fixed in 9a87afa

mikenikles:
Hey Nisar, I see 48px y-space on desktop and 24px on smaller screens which matches the spec. Do you mind pushing a fix for your latest comment?

Could you also do me a favour and push the change needed for the custom breakpoints 🙏?

nisarhassan12:
Thanks, @mikenikles Done. 

> Hey Nisar, I see 48px y-space on desktop and 24px on smaller screens which matches the spec. Do you mind pushing a fix for your latest comment? 

Sorry, my comment was not clear enough It was about the space y for mobile menu items.

mikenikles:
Thanks for wrapping that up Nisar.

PR #682, opened 2021-06-25
hide overflow-x on the body to aviod vertical scaling off the screen.

### LINKED ISSUE #676: Hide the overflow-x on the html element.
On pages where the content overflows horizontally, there appears one can scroll sideways which is not good. We should make sure we have no contents overflowing the viewports and things are fluid regardless of the viewport.

I'm already working on making sure that we don't have contents that overflow the screen at a min-width of 375px in PR https://github.com/gitpod-io/website/pull/657.

Hiding the overflow-x on the HTML element would avoid the following on mobile which currently happens on https://www.gitpod.io/blog/build-a-webhook-for-google-assistant

WhatsApp Image 2021-06-24 at 1 08 04 AM

nisarhassan12:
Fixes #676 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/682"></a>

### DIFF
--- a/src/assets/scss/_reset.scss
+++ b/src/assets/scss/_reset.scss
@@ -12,6 +12,7 @@
 html {
   box-sizing: border-box;
   scroll-behavior: smooth;
+  overflow-x: hidden;
 }
 
 body {


nisarhassan12:
@mikenikles I saw that the body element already had `overflow-x: hidden` set which didn't fix the problem so I had to set this on the HTML element as well.

mikenikles (approved):

PR #697, opened 2021-06-28
Update docs for port visibility

gtsiolis:
Following the updates in https://github.com/gitpod-io/gitpod/pull/4614 and https://github.com/gitpod-io/gitpod/pull/4548, this will update docs for port visibility in the following two sections:

1. [Exposing Ports](https://www.gitpod.io/docs/config-ports)
2. [`.gitpod.yml` Reference](https://www.gitpod.io/docs/references/gitpod-yml)

Cc @corneliusludmann


mikenikles (approved):
Thank you George 🎉

Issue #696, opened 2021-06-28
Provide option to sign up for the newsletter


ChristinFrohne:
**Context:**
We want to provide an option for blog post readers to stay updated about the latest product news. Therefore we'll create a button to easily sign up for the newsletter underneath a blog post. 

The readers shall be able to sign in for the newsletter without logging into Gitpod and without being a registered user. 
[Related internal discussion](https://gitpod.slack.com/archives/C01LDA27WJC/p1624863892182900)

If the email address belongs to a registered Gitpod user, the user notification setting related to "Product tips & inspiration" is turned on and customer.io should be updated accordingly. If the email address does NOT belong to a registered Gitpod user, an entry for that email is upserted in customer.io.

Customer.io should not have, at any time, duplicate emails.

**Next steps:**
- [x] Create Mockup
- [x] Implement the mockup (@nisarhassan12)
- [ ] Technical integration (@jakobhero and @atduarte)
- [ ] Create newsletter segment in customer.io (@ChristinFrohne)

cc @JohannesLandgraf

ChristinFrohne:
Mockup:
<img width="719" alt="Bildschirmfoto 2021-06-28 um 12 44 59" src="https://user-images.githubusercontent.com/46345125/123624286-ad166580-d80e-11eb-9b05-cee4ffe3aa55.png">

nisarhassan12:
>  Implement the mockup @nisarhassan12

I have implemented the mockup in https://github.com/gitpod-io/website/pull/698

ChristinFrohne:
Great, thanks! Could you look into the Segment integration now @jakobhero?

FYI, Pauline (our new college) will publish a blog post on her personal website and she would like to use the code snippet to sign up for the newsletter there as well.

atduarte:
In order to implement this, I believe that the Gitpod app (not the website) should be responsible for the flow from form submission (or API call) to Segment updates. 

1. The website would call the Gitpod app with the email
2. The app would then check whether the email is registered to decide whether to update the associated setting
3. The app publishes a Segment event (to be listened to by customer.io) with the email, user id or anonymous id, and subscription status (i.e. `unsubscribed=false`)
4. Customer data is upserted in Customer.io, using the email as the identifier.

@jakobhero @csweichel thoughts?

jakobhero:
@laushinka is tackling the unsubscription logic in #4761, which will solvee steps 1-3.
[...]

PR #698, opened 2021-06-28
add the ui for the newsletter signup widget.

nisarhassan12:
Addresses part of #696 

This is how it looks:

image

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/698"></a>


src/components/blog/newsletter-signup.svelte
```diff
@@ -0,0 +1,34 @@
...
+</script>
+
+<style type="text/postcss">
+  input {
+    @apply mb-0;
+  }
+
+  button {
+    @apply mr-0 w-32;
+  }
```
mikenikles, line 14: Could these class names be added to the markup below?

src/components/blog/newsletter-signup.svelte
```diff
@@ -0,0 +1,34 @@
+<script>
+  const submitEmail = async () => {
+    // todo submit email
```
mikenikles, line 3: We shouldn't merge this PR until this is implemented, correct?

mikenikles (changes requested):

src/components/blog/newsletter-signup.svelte
nisarhassan12: Yep. I'll mark this Pr as a draft for now.

src/components/blog/newsletter-signup.svelte
nisarhassan12: Thanks. Good Catch Done.

ChristinFrohne:
Thanks, @nisarhassan12 

- [ ] Why did you change the height of the "sign-in" button? It seems quite high right now and it doesn't align with the button height of the other buttons. Most of them are 46px high, but this button is 53px high. Please, check it. 

- [ ] I feel the call-out is a bit too "loud" right now. Could you decrease the font size? H3 instead of H2 and text-medium instead of text-large? Also, the email input field could be a bit less wide, I don't think anyone has such a long email address. 60% width works well. 
Bildschirmfoto 2021-06-29 um 08 31 06

src/components/blog/newsletter-signup.svelte
nisarhassan12: @mikenikles Sorry I have to revert this because of specificity issues adding the class names to markup does not have any impact.

nisarhassan12:
Thanks, @ChristinFrohne I have made the requested changes:

| Before | After |
| - | - |
| image | image  |

I have made the heading and text smaller as you suggested after that change to me the white container looked too big so I have reduced its width + padding which resulted in the decrease of the width for the input as well. Let me know what you think of this change ?

ChristinFrohne:
Looks great, thanks @nisarhassan12! 🎉

Handing it over to @jakobhero for the Segment integration now. 🙏

mikenikles:
Please let me know if I can help with anything. I'd also like to have a quick look before we merge it to understand the integration 🙏.

mikenikles:
Based on some quick research, we need a consent checkbox to be GDPR-compliant.

mikenikles:
🔓 I'm concerned about the current implementation due to privacy & security reasons.

Email addresses are sensitive and we should not use a client-side snippet to submit them. Anyone could take our implementation & Segment write key to add email addresses to our newsletter. FYI, the Segment docs recommend a server-side implementation for sensitive information ([link](https://segment.com/docs/guides/how-to-guides/collect-on-client-or-server/#sensitive-information))

A best practice approach is to performa a double opt-in flow where we ask users to verify their email address before we add them to the newsletter. Once they verify, we use server-side logic to submit the information to Segment.

In addition, we need a consent checkbox to be GDPR compliant based on a bit of research I did.

If the Segment write key in this PR is used anywhere else, please generate a new key so the one from this PR is invalidated.

ChristinFrohne:
Thanks @mikenikles for the legal research. I also did some research and came to the conclusion that this single opt sign-up works fine, as the data is given voluntarily, the user is informed about the impact of sharing the email and the user preformed a clear affirmative action, signifies agreement to the processing of personal data relating to him or her. 

We could consider adding a small footnote linking to the privacy policy like "You may opt-out any time. View our Privacy Policy."
Big companies like Forbes, smashing magazine, etc. do it similarly. 


"Checkboxes are necessary when you are trying to get consent for two separate things, such as a newsletter and advertising. If you don’t need to use checkboxes, it is much better to avoid them." [1]

Sources: 
1) https://www.mailerlite.com/blog/how-to-create-opt-in-forms-that-still-work-under-gdpr
2) https://www.termsfeed.com/blog/gdpr-email-newsletters/

mikenikles:
Nice find Christin 👍. Thanks for looking into that and what you found is a simpler implementation 🚀.

src/components/blog/newsletter-signup.svelte
```diff
@@ -0,0 +1,79 @@
...
+</script>
+
+<style type="text/postcss">
+  input {
+    @apply mb-0;
+  }
+  button {
+    @apply mr-0 w-32;
+  }
+</style>
```
mikenikles, line 45: I see all class names applied to the markup directly, let's remove this style element.

src/utils/analytics.js
```diff
@@ -0,0 +1,80 @@
+(function () {
```
mikenikles, line 1: Let's remove this if we don't use it

src/components/blog/newsletter-signup.svelte
```diff
@@ -0,0 +1,79 @@
...
+  const submitEmail = async () => {
+    isEmailDirty = true;
+    if (!isEmailValid) {
+      return;
+    }
+
+    try {
+      const response = await fetch("/.netlify/functions/newsletter", {
+        method: "post",
+        body: JSON.stringify(data.email.value),
```
mikenikles, line 25: We don't have to use `JSON.stringify` when the body is a string. This adds unnecessary "" around the email address which results in the following value stored in the spreadsheet:

image

mikenikles:
This works in the deployment preview now. The sheet in the spreadsheet didn't have any headers, which resulted in an error when we tried to persist the email address.

src/components/blog/newsletter-signup.svelte
nisarhassan12: Sorry, I should remove them from the markup having them on the markup does not work due to specificity issues due to conflicting styles specified via custom CSS. The following is how it looks without using the style tag which is not what we want:

image

I have remove those classes from the markup now.

mikenikles (approved):

PR #719, opened 2021-07-05
add onboarding flow + general updates.

### LINKED ISSUE #716: Add onboarding flow to /extension-activation + general updates
**Page improvement**
https://www.gitpod.io/extension-activation/ is the 5th most visited page in the last two months (>9.5k page visits). We should use this opportunity in the onboarding flow to teach users about the configuration.

**Changes:**
- New CTA: configure your project
- adding a section that shortly explains the configuration benefits
- small messaging updates
- general layout clean up, e.g. too little spacing

Bildschirmfoto 2021-07-05 um 10 58 23

See [Mockup](https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=1095%3A105)

nisarhassan12:
Fixes #716 

For reviewing the results visit https://deploy-preview-719--gitpod-kumquat.netlify.app/extension-activation

This is how the page looks:

image

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/719"></a>


src/components/extension-activation/config.svelte
```diff
@@ -0,0 +1,27 @@
...
+    >
+      <li>
+        Help Gitpod understand your repository by creating a gitpod.yml file
+      </li>
+      <li>
+        Let Gitpod build your project before you even open it by turning on
+        prebuilds
+      </li>
+    </ol>
+    <a href="/docs/configure" class="btn-conversion">Configure your project</a>
```
nisarhassan12, line 24: I have linked the following button to https://www.gitpod.io/docs/configure commenting this just to make sure that this is correct. 
image

ChristinFrohne:
Thanks! Some comments regarding the design:
- [ ] Please use the updated screenshot that is used in the mockup. 

- [ ] In the mockup I used the same font size and spacing as in /features. But it seems like we updated the specs on the website. Could you use the same specs here, as for the similar section on /features?
Bildschirmfoto 2021-07-07 um 10 54 18

mikenikles (dismissed):

nisarhassan12:
@ChristinFrohne Please have a look when you get time. FYI the following instances all use the same specs as we discussed in our 1:1

image

image

image

ChristinFrohne:
The width of the container is not consistent. On /features both sections (prebuilds and ephemeral dev environments) were 604px wide before but now one section is 700px wide and the other 630px. Please use a consistent width and check if we can keep the former width of 604px. Thanks!


Bildschirmfoto 2021-07-09 um 12 48 10

nisarhassan12:
@ChristinFrohne I have now made sure that the container width stays consistent it is `636` which is not fixed i.e it can vary in accordance to the viewport width but will be the same for all the instances. 

> I'm not sure if all these changes should be in one PR. What about splitting it up. First merging the extension-activation page and then aligning the layout of all pages in another PR.

While it could have been in separate PRs since the changes here are quite coupled and relevant for addressing the following
image

So it made the most sense to do it all in one go because practically it is just a single component as opposed to 3(which was initially the case). 

While doing so I got rid of the custom CSS and migrated the component to use Tailwind.

ChristinFrohne (approved):
Looks good now, thanks!

PR #726, opened 2021-07-06
Segment integration

atduarte:
Integrate Segment, track pages, and track an example action ("homepage_cta_clicked").

 

Out of scope:
* Remove Google Analytics.
* Implement all events of the tracking plan.
* Proxy Segment JS file.
* Newsletter integration.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/726"></a>

### DIFF
--- a/src/components/index/hero.svelte
+++ b/src/components/index/hero.svelte
@@ -146,7 +146,15 @@
     <div class="hero__action">
       <div class="hero__try-now text-small">
         <div>
-          <a href="#get-started" class="btn-conversion">Try Now</a>
+          <a
+            href="#get-started"
+            on:click={() =>
+              window.analytics.track("product_cta_clicked", {
+                context: "hero",
+                destination: "#get-started",
+              })}
+            class="btn-conversion">Try Now</a
+          >
         </div>
         <p>
           Open a workspace.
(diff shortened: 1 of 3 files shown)


nisarhassan12 (changes requested):
Thanks, @atduarte for taking care of this. The Code LGTM 👍🏿 

I tried to test the track event from the CTA that we have on Home and it indeed works I can see the results being reflected at https://app.segment.com/gitpod/sources/website_production/debugger

But it is not the case for subsequent pages being visited. I don't see them reflected in the debugger unless I manually refresh after going on that page. 

### Edit
Please ignore the above the current behaviour is indeed expected because of the following. 

```
if (isProd() && window.analytics?.initialize) {
      // Track subsequent pages
      window.analytics.page();
    }
```

nisarhassan12 (approved):

atduarte:
In fact, we should be consistent and either (1) block all tracking when not in prod, (2) allow all, or (3) use a different write key. Probably 3 is the best option. Wdyt?

nisarhassan12:
> Probably 3 is the best option.

Thanks, @atduarte 3 is the best option in my opinion as well.

nisarhassan12:
I have tested it again and it works on subsequent page views on staging as well.

jakobhero (changes requested):
Please change the event name to 'product_cta_clicked' with the props context and context_url as we settled on in the tracking plan.

jakobhero (approved):
LGTM

PR #730, opened 2021-07-06
add newsletter signup widget to the changelog. 

### LINKED ISSUE #729: Add the newsletter signup widget to the changelog

mikenikles:
Close #729.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/730"></a>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/730"></a>

<a href="https://at-prebuild-page.staging.gitpod-dev.com/#https://github.com/gitpod-io/website/pull/730"></a>


nisarhassan12:
This is how it looks with the latest changes that I made:

image

src/components/blog/newsletter-signup.svelte
```diff
@@ -1,4 +1,5 @@
 <script lang="ts">
+  export let classNames = "";
```
mikenikles, line 2:
```suggestion
  let clazz = "";
  export { clazz as class };
```

☝️ allows us to use this component like regular HTML elements, i.e. `<NewsletterSignup class="mx-auto mt-large mb-huge" />`

src/components/blog/newsletter-signup.svelte
```diff
@@ -30,28 +31,27 @@
...
-        />
-        <button class="btn-primary" type="submit" disabled={isSubmittedOnce}
-          >Sign up</button
-        >
-      </div>
-    {/if}
-  </form>
-</div>
+<form
+  class={`bg-white shadow-normal rounded-2xl max-w-lg py-small px-xx-small sm:px-small ${classNames}`}
```
mikenikles, line 35: With the suggestion above, `classNames` becomes `clazz`

mikenikles (commented):
I think we shouldn't introduce a custom `classNames` property. I suggested a way to use `class`.

src/components/blog/newsletter-signup.svelte
nisarhassan12: Thanks, @mikenikles for the great tip.

nisarhassan12:
This is how it looks now with the latest changes:

image

src/components/blog/newsletter-signup.svelte
```diff
@@ -30,29 +32,30 @@
...
-        >
-      </div>
-    {/if}
-  </form>
-</div>
+<form
+  class={`bg-white shadow-normal rounded-2xl max-w-lg py-small px-xx-small sm:px-small ${clazz} relative`}
+  on:submit|preventDefault={submitEmail}
+  id="newsletter"
+  on:click={(e) => e.stopPropagation()}
```
mikenikles, line 39: Let's use Svelte handler modifiers, see [Svelte docs](https://svelte.dev/docs#on_element_event)

src/components/blog/newsletter-signup.svelte
```diff
@@ -30,29 +32,30 @@
...
+>
+  <slot name="close-button" />
```
mikenikles, line 41: Does this belong to this component? I would argue this is something only the `src/components/modal.svelte` component needs to know about.

Do you think we can refactor that?

src/components/careers/modal.svelte
```diff
@@ -13,31 +15,34 @@
 </style>
 
 <Modal isOpen={!!career} on:close>
-  <h2 class="jobTitle">{career.title}</h2>
-  <p>{@html career.intro}</p>
-  <p>{@html career.paragraphs}</p>
+  <div class="content text-blob" on:click={(e) => e.stopPropagation()}>
```
mikenikles, line 18: Let's use Svelte handler modifiers, see [Svelte docs](https://svelte.dev/docs#on_element_event)

src/components/modal.svelte
```diff
@@ -8,6 +10,10 @@
   let closeEl: HTMLButtonElement;
   let focusedEl: HTMLElement;
 
+  afterUpdate(() => {
+    closeEl = document.querySelector("[data-id='close-modal-button']");
```
mikenikles, line 14: I wonder if this can be refactored together with the `<slot name="close-button" />` I commented on above.

Once the close button belongs to this modal component, we hopefully can use `bind:this` to get a hold of the `closeEl` the way it is done in the main branch currently.

src/routes/careers.svelte
```diff
@@ -211,7 +215,8 @@
 
     <CareerModal
       career={selectedCareer}
-      on:close={() => (selectedCareer = null)}
+      on:close={clearSelectedCareer}
+      handleClose={clearSelectedCareer}
```
mikenikles, line 219: Could you see if we can avoid using a new `handleClose` prop and instead reuse the existing `on:close` event given both do the same thing here?

mikenikles (commented):
Do you know what caused the formatting changes to the following files?
- src/routes/docs/references/gitpod-yml.md
- src/routes/privacy.svelte

nisarhassan12:
> Do you know what caused the formatting changes to the following files?
>    src/routes/docs/references/gitpod-yml.md
>    src/routes/privacy.svelte


I rebased the latest main into this Pr branch because there was a conflict in src/components/blog/newsletter-signup.svelte via the following:

```shell
git checkout main
git pull 
git checkout -
git merge main
```
It happened within the merge commit https://github.com/gitpod-io/website/pull/730/commits/00b11b24be90860cdf149e72c11ee6d9d08e1e7e maybe it was because of the pre-commit hook maybe.

src/components/modal.svelte
```diff
@@ -1,5 +1,6 @@
 <script lang="ts">
-  import { createEventDispatcher, tick } from "svelte";
+  import { createEventDispatcher } from "svelte";
+  export let closeButtonPosition = "right-6 top-6";
```
nisarhassan12, line 3: @mikenikles I could have used the following here instead of `closeButtonPosition` but then for the consumer of the component it could have been quite confusing as the classes belong to a child and not the component itself and also what if we later decide to pass classes into the modal itself ? 

```svelte
let clazz = "";
export { clazz as class };
```

nisarhassan12:
@mikenikles I have addressed your comments. Please have a look again when you get time. Thanks

src/components/blog/newsletter-signup.svelte
```diff
@@ -30,29 +32,29 @@
...
-        />
-        <button class="btn-primary" type="submit" disabled={isSubmittedOnce}
-          >Sign up</button
-        >
-      </div>
-    {/if}
-  </form>
-</div>
+<form
+  class={`bg-white shadow-normal rounded-2xl max-w-lg py-small px-xx-small sm:px-small ${clazz} relative`}
```
mikenikles, line 36: We don't need to turn this into a template literal. Svelte automatically parses `{clazz}` in the markup of a .svelte file.

src/components/modal.svelte
```diff
@@ -30,14 +31,16 @@
 <svelte:window on:keydown={handleKeydown} />
 
 {#if isOpen}
-  <div class="modal" on:click={closeModal}>
-    <div class="content text-blob" on:click={(e) => e.stopPropagation()}>
+  <div class={`modal flex justify-center items-center`} on:click={closeModal}>
```
mikenikles, line 34: We don't need a template literal for this, it's a regular string with class names.

src/components/modal.svelte
```diff
@@ -30,14 +31,16 @@
...
+    <div class="flex justify-center items-center relative">
       <button
+        class={`absolute ${closeButtonPosition} h-3 w-3 z-10`}
```
mikenikles, line 37: This also doesn't need to be a template literal.

src/routes/careers.svelte
```diff
@@ -209,9 +213,6 @@
       {/each}
     </ul>
 
-    <CareerModal
-      career={selectedCareer}
-      on:close={() => (selectedCareer = null)}
-    />
+    <CareerModal career={selectedCareer} on:close={clearSelectedCareer} />
```
mikenikles, line 216: I remember in a previous commit, we had a second prop that used `clearSelectedCareer`. Since this is reverted, what do you think of reverting this too?

In my opinion, the previous code is more readable when the function is inlined as I don't have to find the `clearSelectedCareer` function.

src/components/modal.svelte
mikenikles: Why do you make positioning of the closing button a responsibility of the modal's consumer? Shouldn't each modal be consistent and have the close button positioned at the same place?

src/routes/careers.svelte
nisarhassan12: Thanks. I have reverted this change.

nisarhassan12:
> Why do you make positioning of the closing button a responsibility of the modal's consumer? Shouldn't each modal be consistent and have the close button positioned at the same place?

Thanks. Mike, though we could have both the buttons positioned at the same place the issue is that the newsletter signups popup area is quite small as compared to the popup that we have on /careers where the button is positioned almost `top-16 right-20` and using this value would not look great for the newsletter, in my opinion, see the below screenshots:

Screen Shot 2021-07-12 at 9 27 28 PM

image


Screen Shot 2021-07-12 at 9 20 48 PM

src/components/modal.svelte
mikenikles: Following up on that. Let me know if you'd like to discuss this on a call @nisarhassan12.

nisarhassan12: Sure. Let's do this when you're back at work today.

PR #736, opened 2021-07-07
Use inclusive language.

mikenikles:
A step towards using inclusive language.

If there are other places that need to be addressed, please don't hesitate to inform us.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/736"></a>

### DIFF
--- a/src/routes/blog/gitpodify.md
+++ b/src/routes/blog/gitpodify.md
@@ -63,13 +63,23 @@ Do you wish Gitpod would do more for your project? Then please read on. 👇
 
 ## Table of contents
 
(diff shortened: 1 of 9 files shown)


src/routes/blog/gitpodify.md
```diff
@@ -63,13 +63,23 @@ Do you wish Gitpod would do more for your project? Then please read on. 👇
 
 ## Table of contents
 
+- [Introducing Gitpod](#introducing-gitpod)
```
mikenikles, line 66: Looks like an extension I recently installed keeps the table of contents updated automatically - neat!

PR #762, opened 2021-07-19
implement the segment call for email signup for the newsletter.

nisarhassan12:
<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/762"></a>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/762"></a>

<a href="https://at-prebuild-page.staging.gitpod-dev.com/#https://github.com/gitpod-io/website/pull/762"></a>


nisarhassan12:
André Duarte told me that: 

> problem of doing it in the frontend is that we don’t load segment when people have DoNotTrack activated. And currently, it also gets blocked by ad blockers

So we have to do this via a Netlify function.

src/functions/newsletter/_signup.ts
```diff
@@ -1,15 +1,19 @@
...
+import { nanoid } from 'nanoid';
+
+const analytics = new Analytics('14WvAwP23g2AZddYGxblnp1BbVAErnsy');
 
 export const signup = async (body: string): Promise<Response> => {
-  const isSavedInSheet = await saveEmailInSheet({
-    sheetTitle: "Newsletter - Signups",
-    data: [body],
+  analytics.track({
+    anonymousId: nanoid(),
```
atduarte, line 9: We should not generate a new anonymous id as we want to relate this activity to the rest of the user activity. The anonymous id is stored as a cookie.

Jakob did something similar in the server [here](https://github.com/gitpod-io/gitpod/blob/ea2780a9ce6f91919e5a24221dc4552c43afd1e6/components/server/src/auth/login-completion-handler.ts#L86).

src/functions/newsletter/_signup.ts
```diff
@@ -1,15 +1,19 @@
...
-import saveEmailInSheet from "../feedback/_save-to-spreadsheet";
+import { nanoid } from 'nanoid';
+
+const analytics = new Analytics('14WvAwP23g2AZddYGxblnp1BbVAErnsy');
 
 export const signup = async (body: string): Promise<Response> => {
-  const isSavedInSheet = await saveEmailInSheet({
-    sheetTitle: "Newsletter - Signups",
-    data: [body],
+  analytics.track({
```
atduarte, line 8: @jakobhero also mentioned we should have an identify() call, to associate the email to the anonymous user. We should define it as `email_untrusted` though, as we are not able to confirm that it is indeed their email.

atduarte (changes requested):

src/functions/newsletter/_signup.ts
```diff
@@ -1,15 +1,19 @@
...
+    anonymousId: nanoid(),
+    event: 'Signed Up',
```
mikenikles, line 10: @jakobhero Are you ok with that event name? If I recall we usually use the `signed-up` format 🤔.

atduarte: It should actually be `newsletter_signup` https://docs.google.com/spreadsheets/d/1m_-7uIcwA2fXVP_xv3tvLE1lV_hal8A4h9SBBOyTkck/edit#gid=0

src/functions/newsletter/_signup.ts
```diff
@@ -1,15 +1,19 @@
...
+    email_untrusted: body,
```
atduarte, line 11: What I meant is that a separate identify() call should be done with this property. This event should remain with `email`.

src/functions/newsletter/_signup.ts
mikenikles: Thanks for the link, I see we use `request.cookies.ajs_anonymous_id`. When I look at my Gitpod workspace, I don't see the `ajs_anonymous_id` cookie set, neither do I when I open a workspace.

We can adjust the code to use that cookie, but in order to test I'll need to understand where it gets set. Can you provide instructions on how I can get that cookie set?

atduarte: Ad-blocker maybe? 
image

jakobhero: > We can adjust the code to use that cookie, but in order to test I'll need to understand where it gets set. Can you provide instructions on how I can get that cookie set?
It should get set in the `page()` call when a user visits a gitpod.io site for the first time. For more context, see [here](https://segment.com/docs/guides/how-to-guides/collect-pageviews-serverside/).

JohannesLandgraf:
@mikenikles can we merge, anything still blocking here?

atduarte:
Looking into this (cc @mikenikles)

atduarte:
I was unable to test the netlify functions in the dev environment. 

Up for discussion: whether we want to move the Segment write keys to an env var.

src/functions/newsletter/index.ts
```diff
@@ -1,9 +1,52 @@
...
+
+const identify = promisify(analytics.identify).bind(analytics);
+const track = promisify(analytics.track).bind(analytics);
+
+interface SignupInfo {
+  email: string;
+  anonymousId?: string;
+}
+
+const signup = async (info: SignupInfo): Promise<void> => {
```
atduarte, line 20: @mikenikles is having a separate `_signup` file a convention or it is only when there are multiple of these?

atduarte:
Gitpod app must know whether an existing user subscribed to the newsletter in order to update their preferences.

PR #764, opened 2021-07-19
adjust avatar links + share links.

### LINKED ISSUE #691: [blog]: use social media profiles in the ui when they are defined and not empty.
Currently we are linking to the twitter from the profile widget on the blog posts even if the twitter profile is an empty string see www.gitpod.io/blog/open-vsx-in-gitpod for example eventhough his twitter profile has a value `''` the profile widget still links to twitter.com with `''` at the end. 

This is what the share on twitter link results in We should not add the `@` if the twitter profile is an empty string. 

Search & Install Extensions in Your Gitpod Workspace by @, @svenefftinge https://www.gitpod.io/blog/open-vsx-in-gitpod

nisarhassan12:
Fixes #691 + does a few other things:

Removes the extra `@` if the second or third author does not have a Twitter profile:

image

Set the Avatar links + share links target to `_blank`. 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/764"></a>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/764"></a>

<a href="https://at-prebuild-page.staging.gitpod-dev.com/#https://github.com/gitpod-io/website/pull/764"></a>


src/components/blog/blog-content-layout.svelte
```diff
@@ -32,23 +32,27 @@
...
-  const authorSocialMediaLinks = Object.entries(authors).reduce(
-    (displayNames, [username, profile]) => {
-      displayNames[
-        username
-      ] = `https://twitter.com/${profile.socialProfiles.twitter}`;
-      return displayNames;
-    },
-    {}
-  );
+  const renderTwittwrHandles = () => {
```
mikenikles, line 35: There's a spelling mistake in `renderTwittwrHandles`.

As an exercise, how about you try to rewrite this entire function as a one-liner? The following is a good start and from there, you should be able to use array functions to do the rest, possibly with a single [`reduce()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce).

```js
const renderTwitterHandles = () => author.split(", ").reduce();
```

mikenikles (changes requested):

src/components/blog/blog-content-layout.svelte
nisarhassan12: Thanks @mikenikles I have fixed the spelling mistake and updated the implementation to use a single reduce now it is in theory a one liner with the explicit return but the formatter has made the code span multiple lines. I have also created a helper function + abstracted the `authors.split(", ")` in a variable for use in more then one place.

```ts
const hasATwitterProfile = (authorName) =>
  !!authors[authorName].socialProfiles.twitter;

const writers = author.split(", ");

const renderTwitterHandles = () =>
  writers.reduce(
    (accumlator, currentValue, currentIndex) =>
      accumlator +
      (writers[currentIndex + 1] &&
      hasATwitterProfile(writers[currentIndex + 1])
        ? `${
            hasATwitterProfile(writers[currentIndex])
              ? `@${currentValue}, `
              : ""
          }`
        : `${
            hasATwitterProfile(writers[currentIndex])
              ? `@${currentValue}`
              : ""
          }`),
    ""
  );
```

The previous implementation was quite hard to read and understand and looked quite complex but what we have got with `.reduce` is not satisfying either.

```ts
const renderTwittwrHandles = () => {
  let writers = author.split(", ");
  let result = "";
  for (let i = 0; i < writers.length; i++) {
    if (
      authors[writers[i]].socialProfiles.twitter &&
      authors[writers[i + 1]] &&
      authors[writers[i + 1]].socialProfiles.twitter
    ) {
      result += `@${authors[writers[i]].socialProfiles.twitter}, `;
    } else if (authors[writers[i]].socialProfiles.twitter) {
      result += `@${authors[writers[i]].socialProfiles.twitter}`;
    }
  }

  return result;
};
  ```

mikenikles: As discussed on the call, let's use an author's full name if they don't have a Twitter handle instead of omitting them completely.

Also, to simplify the code, you can concatenate names + , and remove the trailing `, ` at the end of the function.

PR #768, opened 2021-07-20
Update the link indicator. 

nisarhassan12:
Addresses part of [internal slack](https://gitpod.slack.com/archives/C022HAT0H5Z/p1626413868051200). 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/768"></a>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/768"></a>

<a href="https://at-prebuild-page.staging.gitpod-dev.com/#https://github.com/gitpod-io/website/pull/768"></a>


src/assets/global.scss
```diff
@@ -1,3 +1,11 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+a.highlighted {
```
mikenikles, line 5: This file, albeit its unfortunate name, should not be used for anything other than the three existing lines related to Tailwind CSS.

Ideally, I'd prefer this to be encapsulated in a Svelte component. Something like `<NavLink>` or similar we can use for both the header and footer. With that, the highlight-specific styles live in that component and don't have to be globally available when we really only need them in the header and footer.

nisarhassan12: Thanks. Mike. I can make a Svelte for the highlighted link but that would introduce a lot of complexiety i.e in the `NavItem` component the markup looks like this:
 
```svelte
<a
  class:active={isActivePage}
  class:highlighted={isHighlighted}
  {href}
  on:click
  sveltekit:prefetch
  class="relative text-black text-p-large sm:text-dark-grey sm:hover:text-black sm:focus:text-black"
>
  {label}
</a>

```
and in the footer the markup looks like this:

```svelte
<a href="/careers" class="highlighted">Careers</a>
```

looking at the above makes me think we would probably have to conditionally pass/not-pass most of the props depending on where the highlighted component is being used Also we can't extract the indicator into a standalone component because we have to absolutely position it at the top right of the link and for that we have to use `relative` tailwind class on the link to set the correct positioning context unfortunately we can't use margins here to get the desired result. If we go the route of using a standalone indicator component and manually setting the positioning context for the required link then if would not be feasible for consumption into more then two places. 

For using in the nav if we would have to use `{#if}` block to check if `isHighlighted` exists inside of the `each` block and if so then render the Highlighted link other wise the `NavItem` component:

```svelte
<div class="nav-items hidden px-2 space-x-6 items-center md:space-x-12">
      {#each navItems as navItem}
        <NavItem {navItem} on:click={() => ($menuState = !$menuState)} />
      {/each}
</div>
```

Sorry if I missunderstood you. Would it be ok if we keep it the way it currently is and move the css for it to a different file ? Or do have any other implementation of the indicator in mind ? Thanks

PR #786, opened 2021-07-23
Update VS Code extensions example

gtsiolis:
Small update on the [VS Code Extensions](https://www.gitpod.io/docs/vscode-extensions) docs. This will copy the yaml example form the reference docs added in https://github.com/gitpod-io/website/pull/717.

I've avoided copying larger parts to maintain a single source of truth. Additionally, we could also cross-reference the [`.gitpod.yml` reference](https://www.gitpod.io/docs/references/gitpod-yml#vscodeextensions) docs from here.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/786"></a>


mikenikles (approved):
Thank you George!

PR #796, opened 2021-07-28
updated docs to reflect Gitpod's focus on pursuing SOC 2 for now

KlasenK:
the docs still read that we're pursuing ISO 27001 which is currently not the case anymore.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/796"></a>


nisarhassan12 (approved):
Thanks for the PR. It looks good to me.

PR #878, opened 2021-08-24
make the click/tap area for the modal close button bigger for better accessibility

### LINKED ISSUE #811: we should make the close button larger on mobile for better accessibility
I was using the careers modal on mobile and once the focus is lost on the close button via tapping anywhere else I found it hard to close the modal as it is quite small and I have to tap precisely. We should increase the size to maybe 48x48 px as it is recommended tap target size. 

Screen Shot 2021-08-02 at 9 35 47 PM

nisarhassan12:
Fixes #811


nisarhassan12:
@ChristinFrohne I initially tried to make the close button icon bigger but it didn't look good so now instead I increased the clickable area around the image. The clickable area for the button is `40px` instead of `12px`(which we currently have in prod). While reviewing it would make sense to compare this Pr preview with prod.

PR #903, opened 2021-08-30
 Turn canvas and JS based animations off if user has prefers reduced motion set.

### LINKED ISSUE #145: Turn canvas and JS based animations off if user has `prefers reduced motion` set.
On the home page, we should turn off the animation if someone has prefers reduced motion set.

it is a `canvas`. There are two ways i can think of to solve it:  

1. CSS: Set visibility hidden on the canvas with a media query and add a background image to the parent of the canvas.
2. JS: use `matchMedia("(prefers-reduced-motion)");` to see if it's set, and if it is, then make the animation 0s, so you go directly to the end state.

All CSS animations and transitions are already set for prefers reduced motion with a duration of 0s set with !important.

nisarhassan12:
Fixes #145

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/903"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/903"></a>


nisarhassan12:
@mikenikles For the terminal animation on `/` this is how it looks with the preference turned on. So far I'm unable to figure out How can I stop the cursor from blinking the console is built using the Canvas API and I'm not well familiar with that. 

Aug-31-2021 03-21-35

src/utils/helpers.ts
```diff
@@ -18,3 +18,13 @@ export const showHideOverflowY = (bool: boolean) => {
     html.classList.remove("overflow-y-hidden");
   }
 };
+
+export const getPrefersReducedMotion = () => {
```
mikenikles, line 22: This is a great candidate to use a Svelte store, see https://geoffrich.net/posts/svelte-prefers-reduced-motion-store/

nisarhassan12: Thanks. @mikenikles I have switched to use a Store but accessing window from within the store `set` callback seems to fails for us. It is maybe something specific to Sveltekit maybe. Even though I tried `typeof window !== undefined` before accessing the window object. 

Screen Shot 2021-09-02 at 8 27 49 PM

src/components/careers/structured-data.svelte
```diff
@@ -55,4 +55,4 @@
   };
 </script>
 
-{@html `<script type="application/ld+json">${JSON.stringify(structuredData)}</script>`}
+{@html `<script type="application/ld+json" ✂prettier:content✂="JHtKU09OLnN0cmluZ2lmeShzdHJ1Y3R1cmVkRGF0YSl9">{}</script>`}
```
mikenikles, line 58: :thinking: What happened here? That doesn't look quite right.

I have a fix for this in #923. See https://github.com/gitpod-io/website/pull/923/files#diff-a9f4e29225294468331ed3e4d685e41db476903ddfbe8f6539df348cd0f07bbd

mikenikles:
@nisarhassan12 I pushed a few commits to fix the error you mentioned, hide the console cursor and slightly change the workspaces animation when `prefers-reduce-motion` is set to `true`.

src/components/careers/structured-data.svelte
nisarhassan12: I think I tried format document which messed it up I'll revert this change from  the Pr and have a look at #923

nisarhassan12:
Thanks @mikenikles  for fixing the build and making the other changes. 

Regarding stopping the shuffling for the Workspaces component when prefers-reduced-motion is `true` I was wondering whether it would make sense to have 3 workspaces as showing only two looks a bit odd to me. What do you say ?

| Now | Proposed |
| - | - |
| image | image |

mikenikles:
I opted in to display all workspaces. The reason is that during SSR, we render workspaces 0, 1, 2. When we render on the client and detect `prefers-reduced-motion`, we shouldn't exclude workspace 2 as it leads to an awkward re-rendering of workspaces. By opting to display all workspaces, all we do is leave the existing ones from SSR in place and add the remaining workspaces.

PR #923, opened 2021-09-02
Mikenikles/address svelte check 652

### LINKED ISSUE #652: Address `svelte-check` errors & warnings
As pointed out in https://github.com/gitpod-io/website/pull/630#issue-670657481, we are at a point where it makes sense to run `npm run validate` and address the errors and warnings.

Let's also run the above NPM script as part of a `pre-push` git hook, so that we ensure we do not introduce regression.

mikenikles:
Close #652

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/923"></a>


src/components/segment.svelte
```diff
@@ -110,11 +110,9 @@
     analytics.page();
   });
 
-  $: {
+  $: if ($page.path) {
```
mikenikles, line 113: @nisarhassan12 Could you please validate this? I want to make sure we have at least two people make sure it works as expected.

nisarhassan12: Thanks. I have tested it and it works as expected.

nisarhassan12 (approved):

PR #928, opened 2021-09-03
blog(cake): let there be cake

ghuntley:


JohannesLandgraf:
Great write-up @ghuntley, let's merge! Couple of minor things (feel free to ignore)

- I think it would be great to have a picture that shows the cake as a whole at the end
- The pictures are really enormous and disturb sometimes a bit the reading flow (combine Gallery of Cake with the photo from Camisha?)

PR #934, opened 2021-09-06
Add Screenshot for both browser and desktop VS Code to Mediakit.

### LINKED ISSUE #916: Add Screenshot for both browser and desktop VS Code to Mediakit and Brand Assets
One of our larger customers mentioned that on https://www.gitpod.io/media-kit there are only screenshots that look like desktop VS Code. Similar to https://github.com/features/codespaces we should have two screenshots displaying the editor: 
- One for Desktop VS Code
- One for VS Code running in the browser

nisarhassan12:
Fixes #916

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/934"></a>


nisarhassan12:
@ChristinFrohne I see in the mockup that there is no word Download prefixed before the text but I would be in favour of keeping it as it gives context about the button below that says only PNG. What do you think of adding the word download to the button itself ? I don't think it would look good when there are 2 buttons. 

| Mockup | Implementation |
| - | - |
| Screen Shot 2021-09-06 at 5 00 20 PM | image |

Also now all the Screenshots are in PNG format apart from the last one i.e for the Light theme one. Would make sense to convert that into a PNG as well for the sake of consistency ?

ChristinFrohne (approved):
Thanks! I think it's fine the way it is now.

PR #935, opened 2021-09-06
make the font size for the call-out 24px.

### LINKED ISSUE #925: Adjust font size of the call-out section
Can we use 24px instead of 28px for text in the call-out section? We mostly use 24px for text elements. 

E.g. on /features there are already 7 different font styles which seems way too much. Let's not introduce another one.

nisarhassan12:
Fixes #925

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/935"></a>


ChristinFrohne (approved):
Great, thanks!

PR #973, opened 2021-09-14
Add a Dashboard link for logged in users.

### LINKED ISSUE #972: Update the top nav depending on login state
As a logged out user,
I want to see a "Login" and a "Sign up" link
so that I can perform the appropriate action.

As a logged in user,
I want to see a "Dashboard" link
so that I can navigate there easily.

mikenikles:
Close #972.

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -61,7 +61,6 @@ export const handle: import("@sveltejs/kit").Handle = async ({
     ...response,
     headers: {
       ...response.headers,
-      "Set-Cookie": "gitpod-user=loggedIn; Domain=.gitpod.io; Path=/",
       // Avoid clickjacking attacks, see https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html
       "Content-Security-Policy": "frame-ancestors none;",
     },
--- a/src/components/main-nav/dashboard-button.svelte
+++ b/src/components/main-nav/dashboard-button.svelte
@@ -4,12 +4,12 @@
 </script>
 
 <a
-  href="/#get-started"
+  href="https://gitpod.io/"
   on:click={() =>
-    window.analytics.track("signup_clicked", {
+    window.analytics.track("dashboard_clicked", {
       context: "header_button",
     })}
   class="{clazz} flex items-center justify-center rounded-xl bg-black font-bold text-off-white text-sm focus:text-off-white focus:bg-black-hover hover:text-off-white hover:bg-black-hover"
 >
-  Sign Up
+  Dashboard
 </a>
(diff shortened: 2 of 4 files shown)


src/components/main-nav/sign-up-button.svelte
```diff
@@ -6,7 +6,7 @@
 <a
   href="/#get-started"
   on:click={() =>
-    window.analytics.track("dashboard_clicked", {
+    window.analytics.track("signup_clicked", {
```
mikenikles, line 9: @ChristinFrohne or @jakobhero, do we have a `signup_clicked` event? I made that name up given it's the Sign Up button a user clicks on. However, please let me know if you prefer the `dashboard_clicked` value and I'll revert this change.

I also use the `dashboard_clicked` event in the new `dashboard-button.svelte` component introduced in this PR.

src/hooks.ts
```diff
@@ -61,7 +61,6 @@ export const handle: import("@sveltejs/kit").Handle = async ({
     ...response,
     headers: {
       ...response.headers,
-      "Set-Cookie": "gitpod-user=loggedIn; Domain=.gitpod.io; Path=/",
```
mikenikles, line 64: This cookie is set by the product once a user logs in. With this PR, we look for its existence and if it's available, we display a "Dashboard" button in the navigation, otherwise "Login" and "Sign up" buttons.

mikenikles:
@gtsiolis @svenefftinge this PR - as it stands right now - no longer sets the `gitpod-user=loggedIn` cookie since we now rely on i to dynamically display a "Dashboard" button in the website nav.

This impacts the Login screen ([source](https://github.com/gitpod-io/gitpod/blob/main/components/dashboard/src/Login.tsx)).

**Before this PR**
1. User never logged in to Gitpod
2. User visits www.gitpod.io and the `gitpod-user=loggedIn` cookie gets set
3. User visits the login page and is treated as a returning user, i.e. the welcome screen is not shown

**After this PR**
1. User never logged in to Gitpod
2. User visits www.gitpod.io. No cookie gets set
3. User visits the login page and is treated as a new user, i.e. the welcome screen is shown

Are you ok with this change? If not, do you have a suggestion as to how the website can determine whether a user is logged in or not?

svenefftinge:
Hey Mike, new users should not see the welcome screen if they have seen the website before. It is really just there to provide context to those who end up on gitpod from unknown sources (e.g. open-source repos). The website is much better in telling people what Gitpod is, so having an additional message is an unnecessary onboarding step (friction).

src/components/main-nav/sign-up-button.svelte
jakobhero: hey mike, thanks for looping me in on this and excuse the late reply. we have so far used `dashboard_clicked` for all buttons that redirect to the dashboard, including signup. however, it's fine for me to introduce the new event here as long as it's consistently used for signups.

mikenikles: Nice, I like the extra granularity to differentiate between visitors who are not logged in and click the Sign Up button and visitors who are already logged in and click the Dashboard link.

I'll leave it as is and once merge, we'll start to see this new event.

PR #989, opened 2021-09-16
add individual changelog entry pages.

### LINKED ISSUE #681: Add individual pages for each changelog entry
### Problem to solve

Naturally, our [changelog](https://www.gitpod.io/changelog) will grow and we'll have to add pagination at some point. It would be nice to be able to link directly to a changelong entry in a standalone page instead of having to use an anchor link to the heading, like [the last release](https://www.gitpod.io/changelog#faster-incremental-prebuilds), which will not be possible once we add pagination.

### Proposal

Having a permalink to each entry with a structure like the following could suffice:

- `/changelog/2021-05-31-faster-incremental-prebuilds`

TBD (To be discussed)

Cc @mikenikles @JohannesLandgraf

nisarhassan12:
Fixes #681


ChristinFrohne (commented):
Looks good to me, but it would be great if we could add a "back to Changelog" button to the top of the page. As so often, [Vercel](https://vercel.com/changelog/quickly-navigate-the-dashboard-with-shortcuts) is doing it similarly :D

nisarhassan12:
>  if we could add a "back to Changelog" button
@ChristinFrohne done ✅

gtsiolis:
Great work @nisarhassan12! 🏀

Looking forward to seeing this in gitpod.io!

nisarhassan12:
@mikenikles One small nit which I'm unable to fix is that the date and the headline both have different spacings between the defenders and the link see the below video: 

https://user-images.githubusercontent.com/46004116/134032305-5125a1fd-6008-409f-ac01-f8029b1bb6b2.mov

nisarhassan12:
I was able to fix https://github.com/gitpod-io/website/pull/989/#issuecomment-923050962 in https://github.com/gitpod-io/website/pull/989/commits/a65abfc88ba9148cb569c83d2d13cd96d899e611

src/components/changelog/changelog-entry.svelte
```diff
@@ -0,0 +1,71 @@
...
+  export { clazz as class };
+  import { page } from "$app/stores";
+  import type { ChangelogEntry } from "../../types/changelog-entry.type";
+  import { stringToBeautifiedFragment } from "../../utils/helpers";
+  import BackLink from "./back-link.svelte";
+  import ChangelogDate from "./changelog-date.svelte";
+  import ChangelogLink from "./changelog-link.svelte";
+
+  export let entry: ChangelogEntry;
+  export let isAnEntryAlone = $page.path.split("/").length === 3;
```
mikenikles, line 12: Could this be a local variable rather than a prop? I don't think I see it used as a prop anywhere.

src/components/changelog/changelog-entry.svelte
```diff
@@ -0,0 +1,71 @@
...
+        {title}
+      {:else}
+        <ChangelogLink {href}>
+          {title}
+        </ChangelogLink>
+      {/if}
+    </h2>
+    {@html content}
+  </div>
+</div>
```
mikenikles, line 71: This is one of those candidates we discussed back in April when we reviewed the old website's pricing component.

At this point, most lines are wrapped in an `if` (or `!if`) statement and maintaining this is very likely going to be difficult.

What do you think about either creating two separate components or adding the code that's needed at the places where we currently use this component?

src/contents/changelog/2021-02.md
```diff
@@ -1,15 +1,14 @@
 ---
+title: Synchronize Theia user settings with VS Code
 date: 2021-02-28
+image: 2021-02-28.jpg
+alt: VS Code on Gitpod
```
mikenikles, line 5: 👍  Nice work moving this into frontmatter.

src/hooks.ts
```diff
@@ -12,10 +12,12 @@ export const getContext: import("@sveltejs/kit").GetContext = async () => {
 
   const changelogEntries = await Promise.all(
     Object.entries(import.meta.glob("/src/contents/changelog/*.md")).map(
-      async ([, mod]) => {
-        const { default: content, metadata } = await mod();
+      async ([path, page]) => {
+        const { default: content, metadata } = await page();
+        const filename = path.split("/").pop();
```
mikenikles, line 17: I don't think that's used anywhere, is it?

mikenikles (changes requested):

src/components/changelog/changelog-entry.svelte
nisarhassan12: I just took and the code and added it to the places to we currently use it in.

src/hooks.ts
nisarhassan12: Thanks. I have reverted this change.

src/contents/changelog/2021-03.md
```diff
@@ -1,15 +1,14 @@
 ---
+title: 🍊 Dashboard Redesign
 date: 2021-03-31
+image: 2021-03-31.png
```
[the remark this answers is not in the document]
nisarhassan12: Thanks, @gtsiolis for bringing this up. I have addressed this in https://github.com/gitpod-io/website/pull/989/commits/7d7f9bef491cbda0c8814f977c0fd564c3ece034 

Please note that the images are only gonna show up in prod once this PR is merged and deployed. 

For the Descriptions I myself added the excerpt cc @mikenikles

mikenikles (approved):
👍  LGTM.

PR #1026, opened 2021-09-28
Rename twitter-follow-button.svelte to avoid 500

### LINKED ISSUE #843: Client-rendered blog page fails to load when uBlock Origin is in use
* URL: https://www.gitpod.io/blog/workspace-snapshots
* Browser version: Chrome 91.0.4472.77 (Official Build) (64-bit)
* OS: Ubuntu 18.04
* uBlock Origin 1.37.2

The blog post as rendered server-side appears for a fraction of a second, and is then replaced by a client-rendered 500 error. As such, the issue can be worked around by using uBlock to block all scripts on the website. This prevents the error from rendering and keeps the SSR blog post visible.

The issue may be related to the fact that some bundle chunks are blocked from loading:

image

### LINKED ISSUE #1025: [Docs Feedback]: Browser-side 500 errors when using ABP with social blocking list
### Page URL

https://www.gitpod.io/blog/openvscode-server-launch

### User feedback

The [ABP social blocking](https://easylist.to/) easy-list is quite widely used, and filters out requests matching `/twitter-follow`. Unfortunately this breaks the compiled svelte site because one of the chunks matches this string.

Screenshot 2021-09-28 at 21 13 44

A 500 error is shown both when you navigate directly to a blog post, or when you go to the main blog digest page first, and then navigate to any post via the client-side router.

Screenshot 2021-09-28 at 21 15 47

Renaming the svelte file results in a workaround (I will send a PR)

jldec:
Fixes #1025, fixes #843

## How to test
Install ABP extension in Firefox, and enable `Block social media icons tracking`
Then navigate to `/blog/openvscode-server-launch`

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1026"></a>


mikenikles (approved):
Thanks for the fix, this also takes care of #843 where uBlock Origin causes the same issue when an additional filter is enabled.

PR #1040, opened 2021-10-06
Add Tabnine Blog Post

JohannesLandgraf:


phimae (commented):
> This means that using Tabnine in a Gitpod workspace not only removes the friction from setting up and maintaining brittle dev environments, it also makes you faster and more efficient when writing code.

This is confusing as it may seem that removing friction from setting up and maintaining brittle dev environments comes from adding Tabnine. I want to assume that Gitpod Blog readers understand that but let's make it clear anyway.

**Suggestion**:

This means that a Gitpod workspace removes the friction from setting up and maintaining brittle dev environments, while adding Tabnine to your workspace makes you faster and more efficient when writing code.

Edit:

> If you want to add Tabnine to your own Gitpod project, go ahead and install it in **Gitpod** with a single click.
(Gitpod was missing)

PR #1217, opened 2021-11-05
added "new forge" VC to the list

phimae:


package-lock.json
```diff
@@ -24,7 +24,7 @@
         "github-buttons": "^2.19.1",
         "hast-util-heading-rank": "^2.1.0",
         "hastscript": "^7.0.2",
-        "husky": "^7.0.4",
+        "husky": "^7.0.0",
```
mikenikles, line 27: That shouldn't have happened 🤔. Did you use a new Gitpod workspace to make the changes in this PR or reuse an existing one from last time?

phimae: I went to the website repo's `startups.ts` file and launched a workspace from there - so it was a new workspace
What happened was the same like last time regarding the `package-lock.json` 🤷‍♂️

mikenikles: Thanks, I had a look and made sure we're now on version 7.0.4.

mikenikles (approved):

PR #1223, opened 2021-11-08
fix: update wording for workspace deletion

loujaybee:
## Description

Small wording suggestion to make it a bit clearer that workspace deletion is time-based and not based on a number of old workspaces. 

## Related Issue(s)

N/A

## How to test

* Open in Gitpod
* Navigate to: https://www.gitpod.io/docs/life-of-workspace

## Release Notes

N/A

## Documentation

N/A

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1223"></a>

### DIFF
--- a/src/routes/docs/life-of-workspace.md
+++ b/src/routes/docs/life-of-workspace.md
@@ -40,7 +40,7 @@ Not to worry, your changes are safe. You can navigate back to https://gitpod.io,
 
 Old, unused workspaces are automatically deleted.
 
-Unused workspaces are deleted on a schedule. There is no fixed amount of workspaces that are kept by Gitpod. To know exactly how long your past workspaces will be kept, a message is displayed at the top of the [workspaces list](https://gitpod.io/workspaces/) in your dashboard (the exact number of days may change in the future). Restarting a workspace resets the day counter for this particular workspace.
+Unused workspaces are deleted on a schedule. There is no fixed amount of workspaces that are kept by Gitpod. To know exactly how long your past workspaces will be kept, a message is displayed at the top of the [workspaces list](https://gitpod.io/workspaces/) in your dashboard. Restarting a workspace resets the day counter for this particular workspace.
 
 To prevent a workspace from being deleted, you can pin it in your [list of workspaces](https://gitpod.io/workspaces/). Pinned workspaces are kept forever.
 
--- a/src/routes/docs/life-of-workspace.md
+++ b/src/routes/docs/life-of-workspace.md
@@ -38,7 +38,11 @@ Not to worry, your changes are safe. You can navigate back to https://gitpod.io,
 
 ## Garbage Collection
 
-Old, unused workspaces are automatically deleted. To prevent a workspace from being deleted, you can pin it in your [list of workspaces](https://gitpod.io/workspaces/). Pinned workspaces are kept forever. A message at the top of the workspaces list indicates after how many days unused and unpinned workspaces will get collected (the exact number of days may change in the future). Restarting a workspace resets the day counter for this particular workspace.
+Old, unused workspaces are automatically deleted.
+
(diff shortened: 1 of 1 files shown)


src/routes/docs/life-of-workspace.md
```diff
@@ -38,7 +38,11 @@ Not to worry, your changes are safe. You can navigate back to https://gitpod.io,
 
 ## Garbage Collection
 
-Old, unused workspaces are automatically deleted. To prevent a workspace from being deleted, you can pin it in your [list of workspaces](https://gitpod.io/workspaces/). Pinned workspaces are kept forever. A message at the top of the workspaces list indicates after how many days unused and unpinned workspaces will get collected (the exact number of days may change in the future). Restarting a workspace resets the day counter for this particular workspace.
+Old, unused workspaces are automatically deleted.
+
+Unused workspaces are deleted on a schedule. There is no fixed amount of workspaces that are kept by Gitpod. To know exactly how long your past workspaces will be kept, a message is displayed at the top of the [workspaces list](https://gitpod.io/workspaces/) in your dashboard (the exact number of days may change in the future). Restarting a workspace resets the day counter for this particular workspace.
+
+To prevent a workspace from being deleted, you can pin it in your [list of workspaces](https://gitpod.io/workspaces/). Pinned workspaces are kept forever.
```
loujaybee, line 45: * Moved the order around so that pinning comes after explaining how deletion happens. 
* Mention explicitly: "Unused workspaces are deleted on a schedule"
* Mention explicitly: "There is no fixed amount of workspaces that are kept by Gitpod"

src/routes/docs/life-of-workspace.md
```diff
@@ -38,7 +38,11 @@ Not to worry, your changes are safe. You can navigate back to https://gitpod.io,
 
 ## Garbage Collection
 
-Old, unused workspaces are automatically deleted. To prevent a workspace from being deleted, you can pin it in your [list of workspaces](https://gitpod.io/workspaces/). Pinned workspaces are kept forever. A message at the top of the workspaces list indicates after how many days unused and unpinned workspaces will get collected (the exact number of days may change in the future). Restarting a workspace resets the day counter for this particular workspace.
+Old, unused workspaces are automatically deleted.
+
+Unused workspaces are deleted on a schedule. There is no fixed amount of workspaces that are kept by Gitpod. To know exactly how long your past workspaces will be kept, a message is displayed at the top of the [workspaces list](https://gitpod.io/workspaces/) in your dashboard (the exact number of days may change in the future). Restarting a workspace resets the day counter for this particular workspace.
```
loujaybee, line 43: For me "(the exact number of days may change in the future)" brings more questions than it solves, can we provide more context here, e.g. when or why this value might change? It appears we're adding to justify / explain why a fixed value of the days cannot be provided, so can we give the user more context? It's also a little concerning from a user perspective that workspace deletion times might vary/change on an unspecified future date, so I think some more context would help.

loujaybee:
For transparency @mikenikles I created a `team: IDE` label to attach here, I needed this because we're adding PR's like this to the groundwork board, as a note, and to filter via team view the label must be attached to the PR. We might want to consider adding the other teams as labels if they follow a similar workflow, or you might have other ideas. But thought I'd point out and happy to discuss 🙏 🙏 🙏

src/routes/docs/life-of-workspace.md
mikenikles: If we don't have a clear reason we can communicate, let's not confuse things and remove that part. If for some reason the days do actually change in the future, it's the changelog's job to communicate that.

```suggestion
Unused workspaces are deleted on a schedule. There is no fixed amount of workspaces that are kept by Gitpod. To know exactly how long your past workspaces will be kept, a message is displayed at the top of the [workspaces list](https://gitpod.io/workspaces/) in your dashboard. Restarting a workspace resets the day counter for this particular workspace.
```

mikenikles (approved):
Looks good to me. If you agree with my suggestion to remove that confusing part about # of days until workspaces are deleted, please accept the suggestion and go ahead with merging the PR 🎉.

src/routes/docs/life-of-workspace.md
loujaybee: Thanks @mikenikles !

PR #1240, opened 2021-11-12
small updates.

nisarhassan12:


ChristinFrohne (commented):
Please adjust the font size here as well. Otherwise it LGTM
Bildschirmfoto 2021-11-12 um 11 40 17
.

PR #1243, opened 2021-11-12
Redesign of Docs Landing Page

### LINKED ISSUE #1136: [Docs]: Improve the documentation landing page
### Describe your feature

The landing-page for the documenatation should be improved.
Feel free to add some ideas to it @mikenikles 

### How to test

These changes will probably be on the docs-landing page

**Add testing steps when its clear how the page should look like**

Cahllagerfeld:
## Description
Redesign of Docs Landing Page

## Related Issue(s)

Fixes #1136 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1243"></a>


Cahllagerfeld:
Hi @mikenikles 
I just finished the first draft of the mockups, let me know what you think 😄. Maybe Christin needs to answer some of them 😅 
Here are some open topics: 

- [ ] The Icon for the quickstart seems have a transparent margin when exporting it from figma, the other icons work well though. Thats why I took the docs icon for the look and feel in the current version. grafik
- [ ] Should the spacing of the Timline-Items should be fixed, or dyamic adjusted to the text?
- [ ] I took the `excerpt` property of the Changelog-Entry to be displayed in the timeline, imo the full content would be too massive as we dont want to create a second changelog page, In case this is a too short description we need to think of a logic for stripping the `content`-property to our needs, what do you think about this?
- [ ] I put the Changelog Entries into a store, I had problems when running TS-Code in a `.md`-File, this workaround does the job as expected. Should we adjust the other occurences of the Changelog to use the `store` as well, so we dont double the logic?

src/components/docs/landing-page/get-started.ts
```diff
@@ -0,0 +1,25 @@
+import type { GetStartedItem } from "../../../types/get-started.type";
```
Cahllagerfeld, line 1: I didnt know where to put this TS-File, do you have any recommentations?

Cahllagerfeld (commented):
In case there are some issues related to filenames or file-structure let me know please 👍

src/components/docs/landing-page/get-started.ts
mikenikles: Yep, `src/contents` is where we have similar files. I'm working on the PR and will move it.

Cahllagerfeld: awesome, just ping me in discord if you have any questions or need help with something, I'm glad to support you :D

mikenikles:
I replaced the Svelte store with a Svelte context ([docs](https://svelte.dev/docs#setContext), [tutorial](https://svelte.dev/tutorial/context-api)) in 77e90b2.

It's a great use case for context since the changelog entries are only needed in children of `docs/index.md`.

mikenikles:
@ChristinFrohne could you help with the SVG for quickstarts? Also, what do you think of the design in the [preview deployment](https://deploy-preview-1243--gitpod-kumquat.netlify.app/docs)?

ChristinFrohne:
Awesome, looks great. Thanks! ✨ A few comments:
- the buttons are not centered when resizing the screen
Bildschirmfoto 2021-11-29 um 12 15 44

- could you reduce the circles in the timeline from 2rem to 1.5rem? 
- could you add a bit more spacing between the end of the timeline and the text?

Bildschirmfoto 2021-11-29 um 12 41 13

- sorry, I didn't see before now that you have struggles with the icons. The other icons seem outdated as well (the corner radius is too sharp). Here are all three svgs 
[icons-docs-landingpage.zip](https://github.com/gitpod-io/website/files/7617839/icons-docs-landingpage.zip). Not sure why the framing is different. In the [component library](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Website-Component-Library?node-id=133%3A692) they all have the same frame size. Let me know if it still doesn't work and I can further investigate :)

Cahllagerfeld:
Thanks for your feedback (:

- I increased the spacing below the timeline and adjusted the size of the timeline-circles.
- I think the issue with the transparent border aroud the icons occurs in every icon now :(
- The issue with the not-centered button is caused by the button itself. A posibility would be to truncate the text inside the button I guess

src/components/docs/landing-page/get-started-item.svelte
```diff
@@ -7,11 +7,7 @@
   class="text-center bg-off-white rounded-4xl shadow-normal p-8 flex flex-col"
 >
   <div class="flex justify-center">
-    <img
-      class="shadow-normal mt-4 mb-8"
```
mikenikles: > I think the issue with the transparent border aroud the icons occurs in every icon now :(

☝️ was caused by the `shadow-normal` class. The SVG icons themselves already contain shadow and this class added a second shadow around the `` element.

static/svg/docs/landing-page/docs.svg
```diff
@@ -1,25 +1,25 @@
-<svg width="134" height="144" viewBox="0 0 134 144" fill="none" xmlns="http://www.w3.org/2000/svg">
-<g filter="url(#filter0_dd_3226_1362)">
+<svg width="113" height="85" viewBox="0 0 134 144" fill="none" xmlns="http://www.w3.org/2000/svg">
```
mikenikles, line 1: I standardized the `width` and `height` of all SVG elements to match their size in the mockups.

mikenikles:
@ChristinFrohne what device caused the issue with the buttons not being centered?

Looking at various devices at https://www.responsinator.com/?url=https%3A%2F%2Fdeploy-preview-1243--gitpod-kumquat.netlify.app%2Fdocs doesn't show one that causes it.

If you manually resized the browser window, we likely have other places on the site that could look a bit off. In this particular case, there isn't enough space on your screenshot to show three cards next to each other. As Julian suggested, we can either clip the text in the buttons (not preferred) or find a way to dynamically shrink the font size.

Alternatively, we show the stacked layout at a wider screen size. What do you think?

ChristinFrohne:
>@ChristinFrohne what device caused the issue with the buttons not being centered?

I was manually resizing the browser window. I'd say it's okay if it doesn't look perfect then, but the buttons could still be centered. Another solution could be to reduce the padding on smaller widths. Anyway, this shouldn't be a blocker. It can also be addressed in follow-up PR. 

The Quickstart icon is significantly smaller than the other icons. Would be great if that could be aligned. 
I also noticed that the line height of the explaining text is different than usual (31px instead of 29px). We should watch out to be consistent in our styles. Otherwise, this looks good to me :)

src/components/docs/landing-page/timeline/timeline-item.svelte
```diff
@@ -26,7 +26,7 @@
     >
       {timelineItem.title}
     </a>
-    <p class="rounded-2xl text-p-medium leading-relaxed">
+    <p class="rounded-2xl text-p-medium">
```
Cahllagerfeld, line 29: I removed the `leading-relaxed` class to make the line-heights consistent.
I also investigated the line-height-behaviour in general to keep them consistent.
On all mdsvex-pages the following css is applied
```css
.content-docs,
.content-blog {
  line-height: 177.7%;
  ...
}
I think that may differ from pages which arent generated by mdsvex
```

static/svg/docs/landing-page/quickstart.svg
```diff
@@ -0,0 +1,32 @@
+<svg width="113" height="85" viewBox="0 0 194 190" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Cahllagerfeld, line 1: The SVGs are technically all the same height & width now. The difference seems to be the padding from the individual ones. The Quickstart one is having a bigger than the others. 

I dont really know if this can be resolved easily by manipulating the actual sourcefile of the svg.

Cahllagerfeld (commented):
Two small comments

static/svg/docs/landing-page/quickstart.svg
nisarhassan12: @Cahllagerfeld Thanks for working on this ❤️ 

image

I also ran into the same issue in the past with some icons where the box-shadow is part of the exported SVG or image with whitespace(or padding) because of which the icons look smaller despite having the same height and width. A simple solution that I use is that I put `transform: scale(x);` on the icon so that it appears bigger without affecting the layout and elements around it. See the following https://github.com/gitpod-io/website/blob/d500e1467ba4dd25f4a62cdbc233ecfcc019aa3f/src/components/benefit.svelte#L15

An even simpler solution could be to put `transform="scale(1.5)"` directly on the SVG itself but make sure that the SVG is only used in the component where the transform is required(i.e where we are having the above issue) and not re-used in any other place.

Cahllagerfeld: Thanks for your suggestion @nisarhassan12 🙏 . I'll have a look into this

src/contents/docs/landing-page/get-started.ts
```diff
@@ -0,0 +1,34 @@
+import type { GetStartedItem } from "../../../types/get-started.type";
+
+export const getStartedItems: GetStartedItem[] = [
+  {
+    buttonLabel: "Start reading",
+    description: "Learn about Gitpod by reading our docs.",
+    icon: {
+      path: "/svg/docs/landing-page/docs.svg",
+      transform: "scale(1)",
```
Cahllagerfeld, line 9: added transform to the json

src/components/docs/landing-page/get-started-item.svelte
```diff
@@ -0,0 +1,25 @@
...
+  export let getStartedItem: GetStartedItem;
+</script>
+
+<div
+  class="text-center bg-off-white rounded-4xl shadow-normal p-8 flex flex-col"
+>
+  <div class="flex justify-center">
+    <img
+      src={getStartedItem.icon.path}
+      style={`transform: ${getStartedItem.icon.transform}`}
```
Cahllagerfeld, line 12: Transform property from the json gets applied here

Cahllagerfeld (commented):
Adjustments for fixing the sizing issue of the icons

src/components/docs/landing-page/timeline/timeline-item.svelte
mikenikles: Feel free to remove all custom CSS that results in inconsistent design, especially [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) such as 177.7%.

static/svg/docs/landing-page/quickstart.svg
mikenikles: ⚠️ I'd like the SVGs to be exported so they're all the same size rather than workarounds with CSS. @ChristinFrohne I notice in Figma, the icons are all 80x60 in size, but when they're exported they have different sizes. If you open each SVG in a new browser tab and look at them in order, you notice the Quickstart SVG has a larger white padding than the others.

Is there a way we can export the Figma icons in the 80x60 dimensions?

PR #1248, opened 2021-11-15
fix: rss feed for changelog

Cahllagerfeld:
## Description

Fix Rss-Feed for the changelog
## Related Issue(s)

Fixes #

## How to test

call the route `/changelog/rss.xml`

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1248"></a>


src/routes/changelog/rss.xml.ts
```diff
@@ -33,10 +36,26 @@ export const get: import("@sveltejs/kit").RequestHandler = ({ context }) => {
...
     headers: {
       "Cache-Control": `max-age=0, s-max-age=${ttlInMin * 60}`,
       "Content-Type": "application/rss+xml",
     },
   };
 };
+
+const getChangelogs = async (): Promise<ChangelogEntry[]> => {
+  const changelogEntries = await Promise.all(
+    Object.entries(import.meta.glob("/src/contents/changelog/*.md")).map(
```
Cahllagerfeld, line 49: In order to use that `import.meta` I had to adjust the `tsconfig`

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
 	"compilerOptions": {
+		"module": "ES2020",
```
Cahllagerfeld, line 3: See comment on `import.meta`

I'm not qutie sure if this has any side-effects on the `target: es2019` though

src/routes/changelog/rss.xml.ts
```diff
@@ -33,10 +36,26 @@ export const get: import("@sveltejs/kit").RequestHandler = ({ context }) => {
...
-    body: rssDocument,
+    body: rssDocument(logs),
     headers: {
       "Cache-Control": `max-age=0, s-max-age=${ttlInMin * 60}`,
       "Content-Type": "application/rss+xml",
     },
   };
 };
+
+const getChangelogs = async (): Promise<ChangelogEntry[]> => {
```
Cahllagerfeld, line 47: This method is a slightly adjusted version of the changelog-function from `hooks.ts`

Cahllagerfeld (commented):
Just a few comments on what I did and my thoughts

Cahllagerfeld:
I couldnt find a related issue for that, should I open one for consistancy, or is it just a "on-the-fly"-quick-fix :D

src/routes/changelog/rss.xml.ts
```diff
@@ -1,4 +1,8 @@
-export const get: import("@sveltejs/kit").RequestHandler = ({ context }) => {
+import type { ChangelogEntry } from "../../types/changelog-entry.type";
+
+export const get: import("@sveltejs/kit").RequestHandler = async ({
+  locals,
```
mikenikles, line 4: @Cahllagerfeld The `context` was available in the older version of SvelteKit. The concept changed slightly in newer versions and we can now use `locals`. See `hooks.ts` where I refactored it as part of the SvelteKit upgrade.

mikenikles (approved):
Approving this so you can merge when you get up and if you agree with my changes.

src/routes/changelog/rss.xml.ts
```diff
@@ -29,7 +32,8 @@ export const get: import("@sveltejs/kit").RequestHandler = ({ context }) => {
         'src="https://www.gitpod.io/'
       )}']]></description>
     </item>`
-   )}
+     )
+     .join("")}
```
mikenikles, line 36: This removes the `,` we'd get without it between `</item>,<item>`. That was a bug in the old code too but is now fixed.

Cahllagerfeld:
LGTM! 👍

PR #1263, opened 2021-11-18
update links for newsletter signups.

### LINKED ISSUE #1229: Update links for the newsletter signups
**Problem:**
Currently, we are just storing the signups in a Google Sheet but don't use the information. Automatically integrating the data and aligning the information in case the user already exists, is technically quite complex. See also [#6509
](https://github.com/gitpod-io/gitpod/issues/6509)
**Solution:**
We decided to go for an MVC by linking all "sign up of newsletter" buttons to h[ttps://gitpod.io/notifications](https://gitpod.io/notifications) so the users can update the settings themselves. We might iterate on this flow in the future, but for now, this is the simplest implementation.

nisarhassan12:
Fixes #1229


ChristinFrohne (approved):
LGTM

PR #1268, opened 2021-11-19
[epic]: improve responsive design.

nisarhassan12:
Fixes #1253 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1268"></a>

PR #1265, opened 2021-11-18
[section-ides]: update the ide icons so they don't appear to be blurred. 

nisarhassan12:
Addresses https://github.com/gitpod-io/website/issues/155#issuecomment-947228902 

### For the reviewer

Compare the icons on prod with the icons on the deploy preview here and see if the ones in this PR look better and are not blurred as the ones on prod. 

Not sure why but what we currently have on prod does not seem to be blurred for me I tested on Andriod Chrome.

PR #1271, opened 2021-11-22
improve mobile view for the homepage. 

### LINKED ISSUE #1165: Improve mobile view for landing page
Biggest changes:
- adjusted spacing
- [x] new view for switching between different IDEs
- [x] move explaining text before images

Mockup: https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=2427%3A354

nisarhassan12:
Addresses the following from #1165 

- [ ] ~~adjusted spacing~~
- [x] new view for switching between different IDEs
- [x] move explaining text before images

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1271"></a>


nisarhassan12:
@ChristinFrohne Please review this when you get time. Thanks

ChristinFrohne (commented):
Thanks, a few comments:

- [ ] could you reduce the spacing around the text? See also in the mockup

Bildschirmfoto 2021-11-26 um 13 23 30

- [ ] please check if all distances between text and images in the feature list are consistence. It doesn't always seem like it.
Bildschirmfoto 2021-11-26 um 13 28 07

- [ ] does the toggle follow the guidelines for mobile screens? I thought all clickable elements like button need to be at least 48 x48px. That's why I created another mobile toggle, see mockup
- [ ] again clickable elements should be at least 48px. Please, enlarge the buttons so they are at least 48x48px big and cover the whole width (343 px) -> see mockup
- [ ] the logos of the JetBrain IDEs are still blurry
- [ ] some spacing still seems too big (between different sections, cards, etc.). Please check it

Would be great if you have a closer look at the specs used in the mockups. They might not always be exact, but should give you a great orientation. If you disagree with the spacing we can discuss that directly in the mockup. It could save us both some time :)

nisarhassan12:
Thanks, @ChristinFrohne for the review. 
> - [x] some spacing still seems too big (between different sections, cards, etc.). Please check it
> - [x] could you reduce the spacing around the text? See also in the mockup

I pinged you in a comment here https://github.com/gitpod-io/website/issues/1165#issuecomment-975359549 7 days ago that I'll be tackling the spacing related changes in this PR https://github.com/gitpod-io/website/pull/1268 see the deploy preview for that https://deploy-preview-1268--gitpod-kumquat.netlify.app/ the spacing is already reduced there :) 

> - [ ] please check if all distances between text and images in the feature list are consistence. It doesn't always seem like it.

We discussed this earlier like two months on Slack (we also had a call about it when we were doing the landing page revamp). The spacing is consistent the fact that it does not look consistent is explained here https://gitpod.slack.com/archives/D01KRRKB3LJ/p1632732387006600?thread_ts=1632469344.007400&cid=D01KRRKB3LJ for the area you pointed out in the above comment. We tried to improve upon it with the langing page revamp but it is still not prefect everywhere for most of the instances of the Feature component it is fine. 

> - [ ] does the toggle follow the guidelines for mobile screens? I thought all clickable elements like button need to be at least 48 x48px. That's why I created another mobile toggle, see mockup
> - [ ] again clickable elements should be at least 48px. Please, enlarge the buttons so they are at least 48x48px big and cover the whole width (343 px) -> see mockup

It does not follow the guidelines and that is intentional. Yeah, I agree that all clickable elements should be 48x48 for accessibility but it is not easy to achieve for the various different screen sizes and sometimes the design just does not look good for certain elements and changing or coming up with the new designs there isn't quite always possible. 

The current screenshot toggle that has a height of 28px. I don't think that making its height 48px for mobile/tablet size devices would look nice with the current design. Thanks 🙏 that you proposed a new one that would work a lot better while complying with accessibility guidelines.

image

The close button in the banner only has a height and width of 16px by 16px which is quite bad from an accessibility point of view but we can't make it 48x48 for smaller devices as that is never gonna fit into the design well. 

image

The nav toggle button for mobile and tablets has the dimensions 48x24. 24px height is not enough but if we use 48px it will make the size of the top fixed navbar bigger thus occupying more real estate on the already smaller mobile screens. 

image

With the size being 45x45 px my intention was to support mobile devices with width less the 375px there are many devices below that width. 

the buttons overflow the screen on 360px with 48x48 size see the below screenshot:

image

with 45x45 it is still not perfect but works for width >= 360px. Should I make it 48x48 anyway maybe we can break the last one to the next line for smaller screens ? 
 
Today I tested our website on my cousin's phone which does not have a wider screen as my phone does and the main CTA looked broken on that. Our mobile mockups are designed on a canvas with a width 375px I think we should think beyond that as there are smaller devices as well out there FYI some IPhones like 5 and SE have a viewport width of 320px. 

Screen Shot 2021-11-29 at 6 57 32 PM

> - [ ]  the logos of the JetBrain IDEs are still blurry

Should be fixed in https://github.com/gitpod-io/website/pull/1265

nisarhassan12:
I'm merging this further discussions/review comments can be addressed in new issues and PRs cc @ChristinFrohne

ChristinFrohne:
Thanks @nisarhassan12 

>The close button in the banner only has a height and width of 16px by 16px which is quite bad from an accessibility point of view but we can't make it 48x48 for smaller devices as that is never gonna fit into the design well.

What about having an invisible box around that only makes it easier to click these elements? Would be great if ou take inspiration from other best practices. Leaving the final decision to you

>Today I tested our website on my cousin's phone which does not have a wider screen as my phone does and the main CTA looked broken on that. Our mobile mockups are designed on a canvas with a width 375px I think we should think beyond that as there are smaller devices as well out there FYI some IPhones like 5 and SE have a viewport width of 320px.

As far as I can see it works alright besides the main CTA button. Would be great if you could adjust that button to work on smaller screens as well


- [ ] The JetBrains logos look sharp now, but now the VS Code logo is blurry. 
- [ ] The shadow of the buttons are cut on active state
IMG_8092

nisarhassan12:
@ChristinFrohne I have made this PR https://github.com/gitpod-io/website/pull/1330 to address what you mentioned in https://github.com/gitpod-io/website/pull/1271#issuecomment-987983589. I was not able to reproduce the blurry VS Code logo issue but I updated it anyway with an SVG logo downloaded from here https://code.visualstudio.com/brand

Let me know in case if that still looks blurry.

Issue #611, opened 2021-06-11
Add a "On this page" secondary nav for the documentation


mikenikles:
This is inspired by the likes of:
* https://docs.cypress.io/guides/overview/why-cypress
* https://tailwindcss.com/docs/installation

https://www.npmjs.com/package/rehype-toc looks promising.

PR #1289, opened 2021-11-29
Migrate docs to Tailwind CSS

mikenikles:
Relates to #648.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1289"></a>


src/routes/docs/__layout.svelte
```diff
@@ -6,36 +6,11 @@
...
-    @media (min-width: 769px) {
-      @apply w-3/5;
-    }
-  }
-</style>
-
-<div class="docs-layout row">
-  <div class="menu">
+<div class="pb-10 md:flex md:pt-10">
+  <div class="hidden md:block md:w-1/3 md:pt-24 md:pr-8">
```
mikenikles, line 10: I reduced the width from `w-2/5` to `w-1/3` on purpose. This is in preparation for #611.

src/components/banners/base.svelte
```diff
@@ -26,11 +26,7 @@
 </script>
 
 <style lang="postcss">
-  :global(.banner-is-shown) :global(.hero) {
-    @apply mt-small !important;
-  }
-
-  :global(.banner-is-shown) :global(.docs-layout) {
+  :global(.banner-is-shown) :global(main > div:first-child) {
     @apply mt-small;
```
nisarhassan12, line 30: Docs layout is the only place that really needs it from what I have seen, for the other pages they work well without this margin-top apart from the hero on the home page without the margin-top the contents of the hero appear too close to the top nav. This `mt` is not getting applied to the hero because of a specificity issue shown below. Not sure what can be a good solution here apart from using `!important`

Screen Shot 2021-11-30 at 7 27 20 PM

src/components/docs/feedback-widget.svelte
```diff
@@ -45,17 +45,15 @@
...
-    @apply text-h6;
-  }
 </style>
 
 <div class={clazz}>
   <div
     class="bg-white shadow-normal rounded-2xl max-w-md py-small px-xx-small m-auto"
   >
-    <h2 class="h5 mb-6 text-center justify-center w-full">Was this helpful?</h2>
+    <h2 class="text-lg mb-6 text-center justify-center w-full">
```
nisarhassan12, line 54: Small nit: The font size and line-height have changed here they are 18px/28px and on the live site 20px/24px

src/components/banners/base.svelte
mikenikles: Nice catch 👍. This will get automatically resolved when we migrate the `hero.svelte` component to Tailwind CSS. For the time being, I fixed it in 02b1816.

PR #1293, opened 2021-11-30
Migrate the footer to Tailwind CSS

mikenikles:
Relates to #648.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1293"></a>


nisarhassan12 (commented):
Thanks, @mikenikles for looking into it 🙏🏿 . It is LGTM but there is just one minor thing and that is the social icons now have a white rectangular box(barely noticeable) background behind which should not be there. I see that it appears on prod when the social icons hover which should be fixed as well. 

| Prod | Pr Preview |
| - | - | 
| image | image |

mikenikles:
🙏 Thanks for the review Nisar. I removed the white background in 82efb28.

PR #1296, opened 2021-12-01
Add page-level ToC on docs

### LINKED ISSUE #611: Add a "On this page" secondary nav for the documentation

mikenikles:
Close #611.

Tasks
- [x] Remove ToC from changelog (6d11d51)
- [x] Add ToC on blog posts (c3552c1)
- [x] Style ToC on docs & blog posts
- [x] Make the ToC sticky so it scrolls with the content and is always visible. Thanks Julian!

Bugs:
- [x] DevX detail page layout issue with the buttons at the top (fixed in 3bf7a83)
- [x] The mobile top nav menu button is off screen on the Editors docs landing page (https://deploy-preview-1296--gitpod-kumquat.netlify.app/docs/editors) (fixed in f715443)

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1296"></a>


mikenikles:
@ChristinFrohne Could you design the "On this page" nav for me please?

It shows up to the right of all docs pages on large screens. See https://deploy-preview-1296--gitpod-kumquat.netlify.app/docs for example. It currently scrolls down, but will eventually be sticky.

src/assets/markdown-commons.scss
```diff
@@ -249,7 +249,6 @@
   a[href*="http://"]:not([href*="www.gitpod.io"]),
   a[href*="https://"]:not([href*="www.gitpod.io"])
   {
-    white-space: nowrap;
```
mikenikles, line 252: @nisarhassan12 I assume you added that because of the external link svg to make sure it's on the same line as the link, is that right?

Do you remember an example of where that was an issue? Keeping that `nowrap` causes layout issues now that we removed `overflow-hidden: x` from the `html` and `body` tags in this PR.

Also, in prod this is currently broken, see the first link on https://www.gitpod.io/docs/editors in a mobile view.

nisarhassan12: > I assume you added that because of the external link svg to make sure it's on the same line as the link, is that right?

Yep, mike you're right this was the reason that I did this. Without it we broke the external links in the following tables that are in the old languages and framework sections on the docs:

image

I think it is fine to remove it as the link wrapping on the next line is not that big of a problem. 

> Also, in prod this is currently broken, see the first link on https://www.gitpod.io/docs/editors in a mobile view.

So far to avoid this I have been using the following class the `markdown-commons.scss` file which is not ideal and something that should have been avoided be me in the first place. 

image

mikenikles: Awesome thanks for the clarification and background info.

Yep the languages and frameworks content is no longer maintained, so I'm fine with that new line in the tables there.

I'll run a few more tests for this PR and look at the new ToC nav on blog posts, otherwise we're super close to merge this

nisarhassan12: Thanks. Mike sounds good.

PR #1318, opened 2021-12-04
update email address for the sales request.

### LINKED ISSUE #1167: Update email address for sales request
All messages from [https://www.gitpod.io/contact/sales](https://www.gitpod.io/contact/sales) should go to sales@gitpod.io

nisarhassan12:
Fixes #1167 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1318"></a>

@ChristinFrohne Can you verify that submitting the Sales form here https://deploy-preview-1318--gitpod-kumquat.netlify.app/contact/sales does send the email to the `sales@gitpod.io` ? Thanks

### DIFF
--- a/src/functions/submit-form.ts
+++ b/src/functions/submit-form.ts
@@ -60,13 +60,12 @@ async function sendEmail(
 const handler: Handler = function (event, _, callback) {
   console.log(JSON.stringify(event.body));
   const SENDGRID_API_KEY = process.env.SENDGRID_API_KEY || "no-key";
-  const SENDGRID_TO_EMAIL =
-    process.env.SENDGRID_TO_EMAIL || "contact-test@gitpod.io";
-
   const email: Email = JSON.parse(event.body!) as Email;
 
   email.to = {
-    email: SENDGRID_TO_EMAIL,
+    email: email.to
+      ? email.to.email
+      : process.env.SENDGRID_TO_EMAIL || "contact-test@gitpod.io",
     name: "Gitpod",
   };
 
--- a/src/functions/submit-form.ts
+++ b/src/functions/submit-form.ts
@@ -60,12 +60,13 @@ async function sendEmail(
 const handler: Handler = function (event, _, callback) {
   console.log(JSON.stringify(event.body));
   const SENDGRID_API_KEY = process.env.SENDGRID_API_KEY || "no-key";
+  const SENDGRID_TO_EMAIL =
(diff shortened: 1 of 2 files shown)


src/functions/submit-form.ts
```diff
@@ -60,13 +60,12 @@ async function sendEmail(
...
-  const SENDGRID_TO_EMAIL =
-    process.env.SENDGRID_TO_EMAIL || "contact-test@gitpod.io";
-
   const email: Email = JSON.parse(event.body!) as Email;
 
   email.to = {
-    email: SENDGRID_TO_EMAIL,
+    email: email.to
+      ? email.to.email
+      : process.env.SENDGRID_TO_EMAIL || "contact-test@gitpod.io",
```
nisarhassan12, line 68: @mikenikles I'm unable to understand why do we have `contact-test@gitpod.io` here the address won't work, right ? Should I remove it ?

mikenikles: It looks like a fallback in the old code to make sure `email.to` always had a value, likely because the Sendgrid client will throw an exception if it's empty.

The email itself doesn't exist, so that's anyway not an ideal solution. See my comment above, we need to refactor this code.

src/functions/submit-form.ts
```diff
@@ -60,13 +60,12 @@ async function sendEmail(
 const handler: Handler = function (event, _, callback) {
   console.log(JSON.stringify(event.body));
   const SENDGRID_API_KEY = process.env.SENDGRID_API_KEY || "no-key";
-  const SENDGRID_TO_EMAIL =
-    process.env.SENDGRID_TO_EMAIL || "contact-test@gitpod.io";
-
   const email: Email = JSON.parse(event.body!) as Email;
```
mikenikles, line 63: ❗ This is a security issue. We can't accept `to` email addresses from an untrusted HTTP request ❗

Please check your emails to see what kind of damage a malicious person can cause. Also, please revert this code asap so we no longer expose this security risk.

mikenikles (changes requested):

mikenikles:
@nisarhassan12 Let me know if you'd like to discuss an alternative solution. The key factor is to make sure we do not expose the `from` email field via an API.

mikenikles:
I'll work on that.

mikenikles:
I don't know what causes it, but `npm run validate` fails in this PR.

Replaced with #1374.

PR #1319, opened 2021-12-04
enable prerendering for /changelog/[slug].

### LINKED ISSUE #1301: `/changelog/[slug]` is not pre-rendered
The module `<script>` tag needs an `export const prerender = true` for the changelog slug route ([source](https://github.com/gitpod-io/website/blob/main/src/routes/changelog/%5Bslug%5D.svelte)).

nisarhassan12:
Fixes #1301 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1319"></a>

### DIFF
--- a/src/contents/changelog/2021-10-15.md
+++ b/src/contents/changelog/2021-10-15.md
@@ -1,5 +1,5 @@
 ---
-title: Introducing Teams & Projects
+title: Introducing Teams and Projects
 excerpt: Teams & projects - it's now all at your fingertips.
 date: 2021-10-15
 image: 2021-10-15.jpg
--- a/src/routes/changelog/[slug].svelte
+++ b/src/routes/changelog/[slug].svelte
@@ -3,6 +3,8 @@
   import type { ChangelogEntry as ChangelogEntryType } from "../../types/changelog-entry.type";
   import BackLink from "../../components/changelog/back-link.svelte";
 
+  export const prerender = true;
+
   export async function load({ page, session }) {
     const changelogEntry: ChangelogEntryType = session.changelogEntries.find(
       (entry: ChangelogEntryType) =>


src/contents/changelog/2021-10-15.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Introducing Teams & Projects
+title: Introducing Teams and Projects
```
nisarhassan12, line 2: @mikenikles After enabling prerendering for `/changelog/[slug]` the build failed 
<details>
  <summary>See Screenshot</summary>
  
</details>

I was able to fix it by removing the & from the title in the frontmatter. Not sure why `&` in the title caused the issue. Do you know why that happened ? Thanks

Also, should I replace `&` with `and` in other places for consistency in the file and possibly in docs too ? Thanks

mikenikles: Special characters, including `&` are valid. The issue is the `stringToBeautifiedFragment()` function which does not take into account that the `page.params.title` variable is URL encoded.

It returns `introducing-teams-&-projects` which is then compared to `introducing-teams-&amp;-projects`. Since this is not equal, no `changelogEntry` is returned from the `load` function.

mikenikles: @nisarhassan12 Could you have a look at that when you get a minute?

mikenikles: I'll fix this.

mikenikles:
@nisarhassan12 Do you have an idea why this PR and #1318 throw the following error for `npm run validate`?

```
/workspace/website/src/components/index/get-started/index.svelte:30:9
Warn: Unknown at rule @apply (scss)
      @media (max-width: 944px) {
        @apply mb-x-small;
      }

====================================
svelte-check found 0 errors, 1 warning, and 0 hint
```

I can't commit my changes to this PR because of this, so I'll replace the PR with #1375

PR #1334, opened 2021-12-08
fix(/chat): retire invitation link and use public server link

ghuntley:
## Description

Now that Gitpod community is big enough we now appear in the public server listing and invitation links are not needed anymore. Hopefully reduces problems such as this https://twitter.com/gbraad/status/1468479684514316293?s=20

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1334"></a>


gbraad (commented):
This works

PR #1338, opened 2021-12-08
Improve docs for prebuilds and custom Dockerfiles

### LINKED ISSUE #1269: [Docs - Prebuilds and Dockerfiles] Clarify currently implemented behavior
Suggestions for improvement to the docs for configuring prebuilds and Dockerfiles:

https://www.gitpod.io/docs/prebuilds
https://www.gitpod.io/docs/config-docker

### Prebuilds
- [x] Present a minimal .gitpod.yml example for a prebuild (just an init: task)
- [x] Explain that prebuilds only preserve the workspace directory, not the entire gitpod home directory. Mention that installs like `npm install -g` will not be preserved in the prebuild snapshot. Suggest Dockerfiles as another way to pre-install global dependencies.
- [x] Explain that prebuilds happen once per commit and are then shared by workspaces for different users. This means that prebuilds do not include any user-specific environment variables which might include credentials.
- [x] Simplify section about webhooks and how to enable them - link to creating Projects
- [x] Improve the `github: prebuilds:` yaml [example](https://www.gitpod.io/docs/prebuilds#configure-the-github-app) - It is currently too easy to confuse yaml true/false values in the example with the actual defaults in the yaml comments.
- [ ] Can we provide a better use case for having both before: and init: tasks
  - If you want both init and command to run in a specific sub-directory, you need a before.
  - 'Before' tasks are useful for setting up dependencies and tools outside of the workspace directory (e.g. npm install -g) without resorting to a Dockerfile, which comes with additional complexity and maintenance costs e.g. you need to commit a new Dockerfile every time there's an update in any of its dependendencies.
- [ ] What happens if you only have a before task - does that trigger a prebuild?
- [ ] When to use multiple tasks and where do the logs go?
- [ ] Improve the explanation of "incremental prebuilds" - mention threshold where it might help a lot.
- [ ] Nice diagrams with explanations in [this discord thread](https://discord.com/channels/816244985187008514/910616440351129670/910617828594430042)

### Custom Dockerfiles
- [ ] Unlike prebuilds, custom Dockerfile image builds only occur when a change to the Dockerfile is committed.
- [ ] Workspaces will not automatically pick up changes to the base image used in the FROM command in the Dockerfile.
- [ ] Limit on time to run docker build + prebuild 
- [ ] Where do the logs for the docker buld go?

jldec:
## Description
Improvement to the docs for configuring prebuilds and Dockerfiles:

Fixes the initial (most important) parts of #1269
[Preview on Netlify](https://deploy-preview-1338--gitpod-kumquat.netlify.app/docs/prebuilds)

Remaining tasks will be tackled in separate PRs.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1338"></a>


mikenikles (approved):
Awesome improvements, thanks Jürgen!

LGTM 👍

Issue #1393, opened 2021-12-23
Migrate the documentation to `gitpod-io/gitpod`


mikenikles:
Due to a technical limitation with SvelteKit (it didn't support symlinks for routes) at the time we initially developed this website (early April 2021), we migrated the docs from `gitpod-io/gitpod` into the `gitpod-io/website` repo.

The technical limitation no longer exists and it's time to migrate the documentation back in the `gitpod-io/gitpod` main repository. It is important to understand that we only migrate Markdown files. If a Markdown file requires Svelte components, such as the Quickstarts page, these Svelte components continue to live in the `gitpod-io/website` repo.

An initial test was done in #1367.

**Tasks**
- [x] Migrate the following into `src/lib` so we can use `$lib` to import files (#1394)
  - [x] `src/components`
  - [x] `src/contents`
  - [x] `src/stores`
  - [x] `src/types` (re-think this generally, this should be `*.d.ts` files instead
- [x] Move `src/routes/docs` to `$projectRoot/gitpod/docs` and symlink `src/routes/docs` -> `../../gitpod/docs` (#1398)
- [x] Evaluate a dynamic approach to generate the docs menu based on the directory structure within `$projectRoot/gitpod/docs` (#1401)
- [ ] Move all `$projectRoot/gitpod/docs` files to the `docs` directory in `gitpod-io/gitpod`
- [ ] Replace `$projectRoot/gitpod` with the `gitpod-io/gitpod` submodule
- [ ] Write a GitHub action that runs periodically to update the submodule & redeploy the website
  * ❓ Do we want to make this more explicit and require a pull request in the `gitpod-io/website` repo?

PR #1398, opened 2021-12-24
Enable `src/routes/docs` symlink

mikenikles:
Relates to #1393.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1398"></a>


mikenikles:
We may have to revert dc11516. Netlify allows both `/pricing` and `/pricing/` and treats them as different URLs. I've had problems in the past and discussed that with support. I'll need to dig that up, but remember the suggestion didn't work and felt hacky.

PR #1403, opened 2021-12-27
Autogenerate Docs Menu

### LINKED ISSUE #1401: Automatically generate the left-side docs menu
For `/docs/*` URLs, we display a menu on the left-hand side that is currently hard-coded in `src/lib/contents/docs/menu.ts`.

In #1393, we move the documentation to a different repository and updating the `menu.ts` file in the website repo is no longer a convenient workflow. We need to auto-generate the menu based on frontmatter available in the documentation Markdown files.

**Tasks**
- [ ] Remove all unused frontmatter properties in `gitpod/docs/**/*.md`.
- [ ] Add a new `status` frontmatter property to mark menu items as "Beta" or "Soon" (possibly others in the future). See https://www.gitpod.io/docs/editors for how it will be used
- [x] Write a new `handle*` function in `src/hooks.ts` to parse the documentation frontmatter and generate the menu.
- [x] Update the `getSession()` function in `src/hooks.ts` to expose the generated docs menu.
- [ ] Update `gitpod/docs/__layout.svelte` to use the generated docs menu.
  - [ ] Test on mobile
  - [ ] Test on desktop
- [ ] Remove `src/lib/contents/docs/menu`

Cahllagerfeld:
## Description

Auto-Generate Docs Menu

## Related Issue(s)

Fixes #1401 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1403"></a>


gitpod/docs/configure/config-ports.md
```diff
@@ -1,6 +1,7 @@
 ---
 section: configure
 title: Exposing Ports
+status:
```
Cahllagerfeld, line 4: I adjusted the folder structure to match the side-menu. For some reason this seems to break the build.. Does this have to do with the link inside the docs-folder?

src/lib/contents/docs/menu.ts
```diff
@@ -112,3 +116,67 @@ export function getMenuContext(
   }
   return context;
 }
+
+export class MenuService {
```
Cahllagerfeld, line 120: I moved the logic for generating the menu to a service-class

src/lib/contents/docs/menu.ts
```diff
@@ -112,3 +116,67 @@ export function getMenuContext(
...
+  public generateMenu(menuEntries: MenuFrontmatter[]) {
+    let menu: any[] = [];
+    const filtered = menuEntries.filter((el) => el.hasOwnProperty("status")); //TODO remove hasOwnProperty
+
+    const headlines = filtered.filter((element) =>
+      element.path.includes("index")
+    );
+
+    const regulars = filtered.filter(
+      (element) => !element.path.includes("index")
```
Cahllagerfeld, line 132: I'm using the path of the `md` files quite extensively, because they represent the structure of the Menu, I need to properly test this, maybe thats a bit too fragile

Cahllagerfeld (commented):
For some reason the build breaks, this might has something to do with the link inside the docs-folder

gitpod/docs/configure/config-ports.md
Cahllagerfeld: All links from inside the project to the corresponding docs-pages needs to be adjusted...

Cahllagerfeld:
I think the main question is, if we want to restructure the paths be consistent with the structure of the menu...
If not I think I have a other possible solution. What do you think about this?

Cahllagerfeld:
I reverted the path-changes for now, I think the changes would get too big for now

src/lib/contents/docs/menu.ts
mikenikles: I agree, I have an uneasy feeling about that after thinking about it today.

mikenikles:
I thought about that more and came up with an alternative solution in #1420. It takes into account what we discussed on the call, i.e. the docs menu structure.

For now, it only contains two sections for testing purposes.

mikenikles:
Discussed with Julian, closing this and we'll address it in #1547.

PR #1406, opened 2021-12-29
index/section-ides: correct Browser/Desktop label

peat-psuwit:
## Description
The 'toggleChecked' state refers to the right hand side of the toggle.
The code assigns ideType to 'desktop' if toggled, so the right hand side
has to be labeled "Desktop", not "Browser". Vice-versa for the other
side.

## Related Issue(s)

(Didn't file one)

## How to test
Load the website and scroll to "Your environment, your tools, your craft" section. Toggle between Browser and Desktop. The toggle label should correspond to the screenshot below (assuming VS Code is selected by default, the browser one is the one with an address bar).

## Release Notes

N/A?

## Documentation
N/A?

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1406"></a>

### DIFF
--- a/src/lib/components/index/section-ides.svelte
+++ b/src/lib/components/index/section-ides.svelte
@@ -32,8 +32,8 @@
     <div class="max-w-5xl mx-auto">
       <Toggle
         class="mt-x-small mb-macro"
-        labelLeft="Desktop"
-        labelRight="Browser"
+        labelLeft="Browser"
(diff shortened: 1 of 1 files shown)


mikenikles:
Thank you so much! I'm surprised we haven't noticed that sooner given it's such a central part of Gitpod's value proposition 😅.

PR #1420, opened 2022-01-04
1401 Alternative solution

mikenikles:
This solution uses a config file for the docs menu structure.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1420"></a>


src/hooks.ts
```diff
@@ -7,20 +8,25 @@ export const getSession: import("@sveltejs/kit").GetSession = async (
...
-    )
-  );
-  request.locals.docsMenu = menuEntries;
+  const docsFileToFrontmatterMap: DocsFileToFrontmatterMap = {};
+  const docsPages = await import.meta.glob("/src/routes/docs/**/*.md");
+  for (const path in docsPages) {
+    const page = await docsPages[path]();
+    if (
+      page.metadata &&
+      (page.metadata.section === "/" || page.metadata.section === "editors")
```
Cahllagerfeld, line 22: In the current scenario, you just load the Sections `/` and `editors`. Do we need to prefilter here, so only `.md`-files with a valid frontmatter are loaded?

src/lib/contents/docs/menu.ts
```diff
@@ -1,90 +1,66 @@
...
-      (regular) => regular.section === headline.section
-    );
-    children.sort((a, b) => {
-      if (!a.title || !b.title) return;
-      return a.title.localeCompare(b.title);
-    });
-    return children.map((child) => this.mapFrontmatterToMenuentry(child));
-  }
+    menuSection.subMenu = section.items?.map((item) => {
+      const itemFrontmatter =
```
Cahllagerfeld, line 29: Does this line require us to restructure the folder structure?. `Integrations` for example is the scenario that only the `index.md`-file is in the corresponsing folder - The rest of the files are in the `docs`-folder, but on top-level

mikenikles: Good catch, that won't work as is. To be honest, I want to restructure the docs and add redirects to old URLs. We still have many `*.md` files located at the docs root because that's where they used to be for the past two years.

It's baggage I want to get rid of and properly structure the file system so it matches the URL structure. Maybe now is the time to do that. I need a clear brain in the morning to dive into this.

src/hooks.ts
mikenikles: Yeah that was for testing purposes so I didn't have to refactor all docs. We likely only need `if (page.metadata)`.

mikenikles:
Discussed with Julian, closing this and we'll address it in #1547.

PR #1425, opened 2022-01-05
improve the mobile dropdown menu.

### LINKED ISSUE #1006: Improve mobile drop down menu
We should find a way to highlight the "Login" und "Sign-up" buttons in the drop-down menu better. Right now it just looks like another navigation item and not like a CTA. 
Bildschirmfoto 2021-09-20 um 16 26 31

nisarhassan12:
Fixes #1006 

| Before | After |
| - | - |
| image | image |

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1425"></a>


ChristinFrohne (commented):
Cool, thanks!
Outstanding issues:
- [x] the menu stays open when I click on "signup" because of that I can't see the content underneath

- [ ] I couldn't find a way to test the menu when I'm already logged in and the "sign up" button changes to "dashboard". Does that work?

nisarhassan12:
>  - [ ] I couldn't find a way to test the menu when I'm already logged in and the "sign up" button changes to "dashboard". Does that work?

@ChristinFrohne that can only be tested on the prod as the cookie that the Dashboard sets is only accessible there we will have to merge the PR to test that. 

Also, How do you envision the menu to look on Tablet size devices ? i.e we decided to have the Signup CTA next to the menu toggle so that it gets more traction we can't turn that into black as that will have two buttons that are black next to each other(when the menu is not open). 

image

ChristinFrohne:
Can we hide the sign-up button in the main navigation when the menu is open? The sign-up button should then be part of the drop-down menu
Bildschirmfoto 2022-01-11 um 10 02 51

PR #1427, opened 2022-01-05
Use Lite-YT-Embed

### LINKED ISSUE #1175: [Docs - New Feature]: Lighter YouTube Embeds
### Describe your feature

The YouTube Embeds provided by YouTube themselves are extremely slow, I really mean that, even Google themselves suggest using a different one. The one they're suggesting is https://github.com/paulirish/lite-youtube-embed. It's about 200 times faster than the official one. There is basically no downside to this, less stuff is preloaded for the initial load so people who don't intend on playing the video in the first place don't have to pay the cost of all the bandwidth needed for the embed. 

https://github.com/gitpod-io/website/blob/main/src/components/youtube-embed.svelte

### How to test

Open a page like https://www.gitpod.io/docs/getting-started and in the Network tab of the DevTools see the change in bandwidth consumed.

Cahllagerfeld:
closes #1175 

I had a play with it. Let me share my thoughts: 
- The Package provides a web-component which generates several other elements under the hood
- @nisarhassan12 and me paired on this. We had to exchange some of the provided styles, to make the UI fit for our needs. This means, we extracted the corresponding fields and put them to the `assets`. When having the corresponding files in the `assets` we have no need to install the package itself, because nothing get imported from there.
- I added the imports to it like mentioned in the docs, within the `app.html`. This could be moved to a `svelte:head` Tag. By doing so, I had the issue that the app couldn't find the file then and threw an error though. The import references the files from the `assets`
- The Package doesn' t provide event-hooks. This means that we would need to think of another way for tracking the `screencast_started`-event.

We should review this, once you are back from your vacation @mikenikles 
<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1427"></a>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,8 @@
         "@slack/webhook": "^6.0.0",
         "emoji-regex": "^10.0.0",
         "google-spreadsheet": "^3.1.15",
-        "js-cookie": "^3.0.1"
+        "js-cookie": "^3.0.1",
+        "lite-youtube-embed": "^0.2.0"
       },
       "devDependencies": {
         "@jsdevtools/rehype-toc": "^3.0.2",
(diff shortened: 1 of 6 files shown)


Cahllagerfeld:
https://user-images.githubusercontent.com/43843195/148219491-0d9a6d33-df64-4305-881b-237f852d01b5.mov

Because its not a svelte-component. It seems like having an issue with rerendering on prop-change. As shown in the screencast, the player doesn't stop/update the video when switching routes. We would have to think of a way to do this `onDestroy` / `beforeUpdate`

mikenikles:
I built a Svelte port of Paul Irish's `lite-youtube-embed`, but ran into https://github.com/paulirish/lite-youtube-embed/issues/6.

The current implementation of this PR has the same issue. Also, we do need events for analytics. I get the impression this causes more headaches than it solves. Should we close this and wait retry when https://github.com/paulirish/lite-youtube-embed/pull/109 gets merged?

Cahllagerfeld:
I think this is a good approach 👍 Do we need to build a svelte-port for `lite-youtube-embed`, or can we use this is as it is?

mikenikles:
I'd leave the implementation the way it currently is since we can't really figure out a way to make it work _and_ have analytics events as well.

Cahllagerfeld:
Should we keep this PR open then, or close it?

Cahllagerfeld:
Postponed for now

PR #1431, opened 2022-01-07
fix(lang/js): Invalid nodejs version

axonasif:
## Description
The nodejs version `14.08` seems to be nonexistent. Hence causing trouble to the users who are trying to follow this guide:
https://www.gitpod.io/docs/languages/javascript#node-versions

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/7478

## How to test

- Create a Gitpod workspace with a custom dockerfile:
```dockerfile
FROM gitpod/workspace-full:latest

RUN bash -c 'VERSION="14.8.0" \
    && source $HOME/.nvm/nvm.sh && nvm install $VERSION \
    && nvm use $VERSION && nvm alias default $VERSION'

RUN echo "nvm use default &>/dev/null" >> ~/.bashrc.d/51-nvm-fix
```
- Run `node --version`
- Check if it reports `14.8.0` as version.

## Release Notes

```release-note
Use a proper nodejs version in docs/lang/js as example
```

## Documentation

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1431"></a>


gitpod/docs/languages/javascript.md
```diff
@@ -44,17 +44,16 @@ f
...
-RUN bash -c ". .nvm/nvm.sh \
-    && nvm install 14.08 \
-    && nvm use 14.08 \
-    && nvm alias default 14.08"
+RUN bash -c 'VERSION="14.8.0" \
+    && source $HOME/.nvm/nvm.sh && nvm install $VERSION \
+    && nvm use $VERSION && nvm alias default $VERSION'
 
-RUN echo "nvm use default &>/dev/null" >> ~/.bashrc.d/51-nvm-fix
+RUN printf '%s\n' "nvm use default &>/dev/null" >> ~/.bashrc.d/51-nvm-fix
```
mikenikles, line 56: What's the reason to replace `echo`?

axonasif: No special reason, with `printf` you can be just more consistent in terms of formatting the text.

But then again after I pushed the code, it came to my mind that the `echo` is already working, so I was wondering that if I should revert that change. Was hoping that someone would catch this and tell me, so yeah thank you, what should I do in such a situation 🙏? Just reset the commit and force push to the branch? Should I follow the `If it's working then don't fix it` idea?

mikenikles: We use `echo` throughout the documentation and I found only two occurrences of `printf` in a blog post.

To be consistent, let's use `echo` here as well. Reading the `echo` statement is also simpler since not everyone may be familiar with the `printf` arguments such as `%s\n`.

I wouldn't reset the commit because that also reverts the other changes you made. I'd change `printf` back to `echo` in a new commit.

axonasif: Makes sense, I never thought about that 😅 
I soft-reset the commit, reverted that line and comitted again.
Thank you very much!

mikenikles (approved):

Cahllagerfeld:
@axonasif This can be merged as its approved. Do you want to merge it yourself, or shall I merge it?

Cahllagerfeld:
We usually use `squash` :D

axonasif:
Okay thanks @Cahllagerfeld!

PR #1434, opened 2022-01-07
Create Reusable Comparison-Table

### LINKED ISSUE #1384: create a reusable features table component.
In https://github.com/gitpod-io/website/pull/1084/commits/d32cc8c2ba6c165d04218d3647e44ad3a7ae7b50 I refactored some of the code relevant to the feature tables that we use but not all. Currently, there are some duplications/inconsistencies from page to page that is why we should create a re-usable component that results in consistency everywhere and is easier to re-use. 

Mockup to follow: https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=3012%3A1274

Feedback from Christin that can be addressed as part of this Pr. 

- [ ] ~~We should also improve the table view for tablet size devices as suggested here https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=3541%3A3413 by Christin.~~
- [ ] the logos are overlapping when resizing the window this can be re-produced with the table on https://www.gitpod.io/vs/jetbrains-space
       <details>
        <summary>Screenshot:</summary>
        Screen
       </details>
- [ ] the spacing between the category items column and the card like columns should be decreased.

Cahllagerfeld:
closes #1384

- [x] add option for highlighting a specific column
- [x] Gitpod vs. Codespaces
- [ ] local vs. gitpod
- [ ] option for more than two columns?
- [ ] dark indicator per line?

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1434"></a>


src/lib/components/tables/details.svelte
```diff
@@ -63,7 +63,10 @@
 </dl>
 
 {#if link}
-  <div class="relative h-10 w-full">
+  <div
+    class="relative w-full flex justify-center items-center"
+    style="height: 60px;"
```
Cahllagerfeld, line 68: we need to tackle this when removing all the custom-css

src/lib/assets/scss/_feature_table.scss
```diff
@@ -13,7 +13,9 @@
 .featureTable dd,
 .featureTable dt,
 .featureTable h2 {
-  text-align: center;
+  display: flex;
```
Cahllagerfeld, line 16: this needs to get removed, when removing all custom-css

Cahllagerfeld (commented):
Two comments regarding Custom-Css

Cahllagerfeld:
Technically the feature-table is a grid with 3 columns. In case that this needs to provide a dynamic amount of columns, we need to think of a logic that changes that layout.

Cahllagerfeld:
The Comparison-Table should Support 3-5 Columns

src/lib/components/tables/comparison-table.svelte
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  .toc li {
+    @apply flex items-center pl-0 text-left font-bold;
+  }
+</style>
+
+<div
+  class="featureTable"
+  style="grid-template-columns: repeat({gridCols}, minmax(0, 1fr));"
```
Cahllagerfeld, line 57: **Dirty Hack - Reevaluate before merge**
Done as inline-styling due to override the custom css

Cahllagerfeld:
closed due to tailwind-migration

Issue #1303, opened 2021-12-02
After navigating pages from the footer the back button does not takes the user to the correct position.  


nisarhassan12:
After navigating to different pages of the site via the footer and then navigating back to the homepage via the browser back button does not always takes you to the correct position(from which the user navigated to the page). See the below video to better understand this

https://user-images.githubusercontent.com/46004116/144517427-f4a39333-d09d-4de4-acb2-eba23d4c4467.mov

I was able to reproduce this issue on many pages i.e Pricing, Careers, Local development and so on. Also, I noticed that this does not happen on the Changelog page.

nisarhassan12:
@mikenikles This seems to be impacting the navigation in Safari and the impact is not so small. If one goes to page like Contact or the Changelog and then clicks the Signup button from the fixed nav it does not goes to the Get started now section as expected.  

https://user-images.githubusercontent.com/46004116/145487850-bcb83d1a-4a83-4eb1-84dc-88bdcdc86fd8.mov

Siddhant-K-code:
@chrifro, It is resolved now. See 👀 



https://user-images.githubusercontent.com/55068936/215393058-cc978357-f8d9-4114-8136-7c75eee08d3a.mov

PR #1438, opened 2022-01-07
Update the Screencasts. 

### LINKED ISSUE #1283: Update /screencasts and add a screencast index page
We want to improve the ratio of users visiting /screencasts and actually watching a screencast. As most users click on the first screencast, we decided to highlight that and thereby ease the choices for the user. Furthermore, we want to highlight all existing screencasts on our website and not lose users on YouTube. That's why we created the screencast index page.

[Mockup:](https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=2774%3A650
New Screencasts (1)
)
Screencast-index (2)

Note: there are a few comments in the Figma mockup, helping to understand the mockup

nisarhassan12:
Fixes #1283

Deploy Preview: https://deploy-preview-1438--gitpod-kumquat.netlify.app/screencasts/getting-started-with-gitpod

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1438"></a>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1438"></a>


mikenikles:
@nisarhassan12 is this still a draft or is it ready for review?

nisarhassan12:
@mikenikles it is still a draft. I'm planning to fix this https://github.com/gitpod-io/website/issues/1283 as par of this PR and there is still a lot of stuff remaining.

Cahllagerfeld:
grafik
It looks like the duration isn't set correctly. (At least for me) the duration always stays at 2mins although the actual video length differs 🤔

Cahllagerfeld:
On `screencast`, the last one with `local-companion` contains `undefined`
grafik

nisarhassan12:
@pawlean at the moment below each of the individual screencasts pages we are displaying up to 6 screencasts in the **Next up...** section. What do you think about it? i.e is it fine like this ? or do we want to reduce them to up to 3 ? or have 6 on the first one and 3 on the rest ? or something else ? 
image

ChristinFrohne (commented):
Cool, thanks @nisarhassan12 ✨

Some comments:

- [x] could you adjust the size of the YouTube embed so there is no black margin when watching the video? (the current height is fine, just the width needs to be reduced)

- [ ] after clicking on "view next screenshot" the next video should be in the focus of the screen 

- [x]  please update the text of this section to the following
Bildschirmfoto 2022-01-17 um 11 12 39

- [x] something seems to be off in the "get started" call out. The images are outdated (the colors of the workspaces differ)

- [ ] on mobile: could you reduce the numbers of videos listed under "next up" to 3?

- [ ] on the index page: if the search doesn't find anything, let's display an error message "Sorry, we couldn't find any results matching your search."
- [ ] on the index page:  let's clear the search bar if a suggested topic is selected

- [ ] on the index page: to disable a suggested topic the user should need to click on the topic again. Currently, the results are cleared by clicking anywhere on the page which can be confusing
- [ ] currently, the text underneath the headline is a duplicate of the description underneath each video. Would there be anything else noteworthy to mention in the video description or should be remove the description underneath the video @pawlean ?

nisarhassan12:
> could you adjust the size of the YouTube embed so there is no black margin when watching the video? (the current height is fine, just the width needs to be reduced)

@ChristinFrohne I have adjusted it will work totally fine for desktops but as we get to smaller screens it won't be the case there will be some small black margins.

nisarhassan12:
> - [ ]  after clicking on "view next screenshot" the next video should be in the focus of the screen

@mikenikles I don't know how this can be achieved at the moment for some reason it does not scroll to the top of the next page after the navigation. I have seen other issues where the scroll does not work reliably and the navigation does not start from where it is expected to be #1303 specifies a few.

pawlean:
@nisarhassan12 Can we include this screencast for OpenVSCode (https://www.youtube.com/watch?v=qGR7rgqjdiY)? It's not set up like the other ones (e.g. thumbnail) but it was a bit different compared to the others as it's not Gitpod product related.

pawlean (changes requested):
Apart from the link I mentioned, this is looking sharp. Thanks Nisar!

mikenikles:
> after clicking on "view next screenshot" the next video should be in the focus of the screen

@nisarhassan12 I fixed ☝️ in e866d81. When we navigate from one screencast to another, we use the same component and simply change the content. This means, Svelte does not re-render the component hence the default browser behaviour of scrolling to the top doesn't apply. The `afterNavigate` lifecycle function provided by SvelteKit's `$app/navigate` module lets us solve this by manually scrolling to the top.

PR #1441, opened 2022-01-09
Update the installation docs for helm deprecation.

metcalfc:
Folks are still finding the old docs and following the v0.10.0
installation. We should point them to the new installer and mark the
old method deprecated. Eventually we'll update the docs for a full
installer path and remove the helm path entirely.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1441"></a>


mikenikles:
Hey @metcalfc, this PR breaks SEO scores for the existing Gitpod self-hosted docs. It sounds like we're ok with that since the content at these URLs is deprecated, but I wanted to double check before we merge the PR.

Also, the build breaks because we have a number of internal links that point to documentation that no longer exists now. If you're ok with removing these pages, I'll go ahead and update all other places that point to deprecated docs.

metcalfc:
Thats a good point. I can add redirects for those latest pages to come back to the index for now? Eventually, we'll have more docs for the latest path and some of these could come back in. Would that work?

gitpod/docs/self-hosted/latest/index.md
```diff
@@ -9,12 +9,83 @@ title: Gitpod Self-Hosted
...
-- [Configuring Gitpod Self-Hosted](./latest/configuration)
-- [Administrating Gitpod Self-Hosted](./latest/administration)
-- [Troubleshooting Gitpod Self-Hosted](./latest/troubleshooting)
-- [Updating Gitpod Self-Hosted](./latest/updating)
-- [Gitpod Self-Hosted releases](./latest/releases)
+Gitpod, just as you know it from [gitpod.io](https://gitpod.io), can be deployed and operated on your own infrastructure. It supports different cloud providers, vanilla Kubernetes clusters, and behind the corporate firewalls.
+
+# Installation via Helm (deprecated)
+
+Installing Gitpod via Helm is deprecated and the charts will evetually be removed. If you are looking for the previous documentation you can find it [here](../deprecated/index).
```
atduarte, line 16: eventually*

gitpod/docs/self-hosted/latest/index.md
```diff
@@ -9,12 +9,83 @@ title: Gitpod Self-Hosted
...
+
+# Installation via the new installer
+
+## Supported Kubernetes distributions
+
+Gitpod Self-Hosted runs well on:
+
+- [Amazon Elastic Kubernetes Service](https://github.com/gitpod-io/gitpod-eks-guide)
```
atduarte, line 24: This Guide doesn't use the installer. 🤔

gitpod/docs/self-hosted/latest/index.md
```diff
@@ -9,12 +9,83 @@ title: Gitpod Self-Hosted
...
+- [Google Kubernetes Engine](https://github.com/gitpod-io/gitpod-gke-guide)
+- [Microsoft Azure Kubernetes Service](https://github.com/gitpod-io/gitpod-microsoft-aks-guide)
+- [Kubernetes](https://github.com/gitpod-io/gitpod/tree/main/installer)
```
atduarte, line 27: Would it be better to link directly to https://github.com/gitpod-io/gitpod/blob/main/installer/README.md ?

mikenikles:
@metcalfc Yep that should do the trick. FYI, redirects are configured in `_redirects`.

atduarte:
Closed in favour of https://github.com/gitpod-io/website/pull/1536

PR #1451, opened 2022-01-11
Style: Tooltip from figma

### LINKED ISSUE #1448: improve the tooltips as per the specs.
The specs can be found here: https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Website-Component-Library?node-id=518%3A1074

Relates to https://github.com/gitpod-io/website/pull/1415#pullrequestreview-848884744

Cahllagerfeld:
closes #1448

implementation using `popper.js`

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1451"></a>


mikenikles:
The UI looks good and works as expect. There's an issue when a tooltip has a link as the tooltip disappears when the mouse moves and the link can't be clicked.

See "Encrypted Backups" and "Multi-IDE Support" on the pricing page for example. Any chance popper.js supports a way to keep the tooltip visible when someone clicks on the `?`?

src/lib/components/tooltip.svelte
```diff
@@ -1,73 +1,51 @@
...
+  .arrow::before {
+    @apply visible;
+    content: "";
+    transform: rotate(45deg);
   }
 </style>
 
-<span class="relative" on:mouseleave={mouseLeave}>
+<span
+  on:mouseleave={() => {
```
Cahllagerfeld, line 43: >
>The UI looks good and works as expect. There's an issue when a tooltip has a link as the tooltip disappears when the mouse moves and the link can't be clicked.
See "Encrypted Backups" and "Multi-IDE Support" on the pricing page for example. Any chance popper.js supports a way to keep the tooltip visible when someone clicks on the ??
>
I think the issue is related to that. Whenever the cursor moves out of the span, the `mouseleave` gets triggered. I tested it with removing this event and it works perfectly fine.

nisarhassan12:
@Cahllagerfeld When you get time can you please look into adjusting the tooltip peak arrow position when the bubble is to the left side? 

Screen Shot 2022-01-19 at 1 38 11 PM

Cahllagerfeld:
>Also at the moment, it is quite hard to get to the link Can we use an invisible element below the bubble like we did here  

The Question-Mark can be clicked so it stays open. I'll add the element below it though

>When you get time can you please look into adjusting the tooltip peak arrow position when the bubble is to the left side?

Sure, I'll fix this 👍

ChristinFrohne:
I agree would be great if the links also work on hover. 

Thanks for fixing this. Generally, the tooltips look great! ✨

PR #1469, opened 2022-01-13
Ignore Quickstarts for SEO

### LINKED ISSUE #1462: SEO: Reduce similar content on indexable URLs (Quickstart)
**Issue:** 
Currently, there are URLs that have the same or very similar content as at least one other indexable URL. When this type of duplication occurs, it often has far-reaching consequences for SEO. Duplicate content can lead to poorer performance and a downgrade by Google's algorithm (e.g. "Google Panda"). Therefore the amount of unique content should be increased and the amount of duplicate content should be reduced.
This only affects the Quickstart examples

See all affected URLs: https://docs.google.com/spreadsheets/d/1jpE4tUT8X3HzPF1WtCiYfQwxR0sgZpEBKyqEkKvhuBA/edit#gid=0

**Solution:** 
Option 1: we adjust the content for each quickstart to be more unique
Option 2: we only make /quickstart visible for search engines and hide each quickstart subpage. [[1](https://www.sistrix.com/ask-sistrix/google-index-google-bot-crawler/how-can-i-remove-a-url-on-my-website-from-the-google-index)]

I'd suggest we go for option 2.

Cahllagerfeld:
closes #1462 
would this be a propper way to implement this @mikenikles ?

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1469"></a>


gitpod/docs/quickstart/agda.md
```diff
@@ -7,6 +7,15 @@ title: Agda template
...
 </script>
 
+<script>
+
+  import SEO from "$lib/components/seo.svelte";
+  </script>
+
+<SEO>
+  <meta name="robots" content="noindex" />
+</SEO>
```
mikenikles, line 17: What about this?

```suggestion
<svelte:head>
  <meta name="robots" content="noindex" />
</svelte:head>
```

With that, we don't need the `<script>` tag and since the current `seo.svelte` component only provides a `<svelte:head>` wrapper, that'll save us a bunch of lines we need to add to each quickstart file.

Cahllagerfeld:
I can reach the following results with Lighthouse-SEO for the preview page: 
|mobile|desktop |
|---|---|
| 91  | 91  |

gitpod/docs/quickstart/agda.md
mikenikles: Thanks for the link, time to move on from that approach 😄.

One idea, rather than updating each md file individually, what about a new `gitpod/docs/quickstart/__layout.svelte` with the following content:

```svelte
<script lang="ts">
  import { page } from "$app/stores";
</script>

<svelte:head>
  {#if !$page.url.pathname.endsWith("quickstart")}
    <meta name="robots" content="noindex" />
  {/if}
</svelte:head>

<slot />
```

Either way is fine with me and we can merge this PR.

mikenikles (approved):

Cahllagerfeld:
Decided to move it to a `layout`, its more generic and easier to maintain I think

PR #1470, opened 2022-01-13
Remove `_reset.scss`

### LINKED ISSUE #1453: Remove `_reset.scss`
We need to get rid of this file, for the migratio to tailwind. The styles defined in this Stylesheet affects all other components as we have `*`-Selectors in it.

Cahllagerfeld:
closes #1453

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1470"></a>


Cahllagerfeld:
I think we should discuss if we want to keep all the resets (partially that arent resets anymore, but stylings) and remove / move them to other layers

src/lib/assets/global.css
```diff
@@ -0,0 +1,80 @@
...
+  a:focus {
+    @apply text-black underline decoration-transparent;
+  }
+
+  a[class] {
+    @apply no-underline;
+  }
+
+  ul:where([class]),
+  ol:where([class]) {
```
Cahllagerfeld, line 43: this

src/lib/assets/global.css
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  body {
+    font-family: var(--font);
+    text-rendering: optimizeSpeed;
+    line-height: 1.5;
+    @apply m-0 min-h-screen text-dark-grey bg-sand-light text-p-medium;
+  }
+
+  a {
```
Cahllagerfeld, line 30: I feel like thats a default-styling. What do you think about this? When removing this, we probably need to take a look at EVERY other anchor-tag in the repo

Cahllagerfeld (commented):
Two things I'm not sure about

src/lib/assets/global.css
mikenikles: I agree, we'll have to do a thorough review. For now, I'd say lets merge it as is and look at it as part of the UI library creation.

src/lib/assets/global.css
mikenikles: 👍 I can't think of a better solution until we have a `<List />` component in the UI library.

mikenikles (approved):
👍 2 `scss` files gone 💣

PR #1481, opened 2022-01-17
Remove `_avatars.scss`

Cahllagerfeld:
context: #648

removes `_avatars.scss`


src/lib/components/avatar.svelte
```diff
@@ -3,6 +3,14 @@
   export let socialMediaImgClasses;
 </script>
 
+<style lang="postcss">
+  .linkless {
```
Cahllagerfeld, line 7: the `.linkless` class is used only once in the project. Thats why I didn't scope it to the `.avatars` parent. If thats needed, let me know 👍

mikenikles: That's fine, styles in Svelte components are scoped to that one component by default. You could even use:

```css
img {...}
```

Having said that, I'm dropping it entirely in favour of a negative spacing class on the parent element.

PR #1485, opened 2022-01-18
Remove `_teams.scss`

Cahllagerfeld:
context: #648

removes _teams.scss

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1485"></a>


Cahllagerfeld:
There's an issue with the buttons on the `/about` page for example. They are affected by custom-stylings from "button.scss". We probably need to replace them when migrating the buttons

mikenikles:
I fixed the buttons in 249dc6c for the `/about` page and the one blog post that used them too. I opted for a separate component to isolate anything related to them. They're not listed in our UI library and I personally think we should get rid of them. To avoid scope creep, let's keep it in that separate component and address after the migration.

PR #1495, opened 2022-01-19
Adjust Tooltip

Cahllagerfeld:
adjust position of `arrow` when displayed to the left

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1495"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1495"></a>


Cahllagerfeld:
grafik
The `span`-wrapper is in the coding. The Problem is that theres a small border between the root-element and the actual tooltip. Thats why the `mouseleave`-event fires. It only works when moving the cursor to the `arrow` and then to the tooltip. We need to extend the actual `mousearea` of the `root`-Element or the `tooltip` itself. Do you have any idea for this @nisarhassan12 ? I had a look at `::before` or `::after` but didnt get it working :(

nisarhassan12:
@Cahllagerfeld I'll play around with it and push a fix here if it worked.

Cahllagerfeld:
> @Cahllagerfeld I'll play around with it and push a fix here if it worked.

awesome, thank you 🙏🏼

src/lib/components/tooltip.svelte
```diff
@@ -58,17 +73,14 @@
...
-        isClicked = !isClicked;
-      }}
       use:popperRef
       src="/svg/question-mark.svg"
       alt="Tooltip"
       class="h-5 w-5"
     />
   </button>
 
-  {#if isClicked || isRendered}
```
Cahllagerfeld, line 71: See [this](https://github.com/gitpod-io/website/pull/1451#issuecomment-1012557963) comment here from Mike. I dont know if we want to delete the `click`-Feature. What do you think about this?

Cahllagerfeld: The remaining changes look good to me 🚀

nisarhassan12: Thanks, Julian. I don't think the click is needed anymore i.e mouse users can now easily use the links by hovering via a mouse and keyboard users can use tab and mobile users can tap because we have `on:focus` and `on:blur` in place. What do you say? I'm fine if we keep it :) 

If I'm not wrong I think @mikenikles asked for it as a way for the users to be able to use the links inside of the tooltips.

nisarhassan12:
@Cahllagerfeld I see a small issue that I was not able to fix and that has to do with the arrow tip being overlapping with the content see the below screenshot (it slightly covers the link underline).

Screen Shot 2022-01-19 at 6 13 55 PM

I tried to search for a solution but I didn't find one I tried putting a lower negative `z-index` but that can't work because of the way popper works. Does anything come up in your mind ? maybe it would be against the Figma specs but what if we translate the arrow by 1 or 2 pixels to avoid the overlapping ? Thanks

nisarhassan12 (approved):

PR #1496, opened 2022-01-19
make it easier to work with images.

nisarhassan12:
Addresses https://gitpod.slack.com/archives/C01KGM9ETU6/p1642584924030100

| Before | After |
| - | - |
| image | image |

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1496"></a>


src/lib/assets/global.css
```diff
@@ -27,6 +27,21 @@
...
+  img,
+  picture {
+    @apply max-w-full block object-contain;
+  }
+
+  /* Ensure accessibility. Remove outline if we may miss images in prod */
+  img:not([alt]) {
+    filter: blur(10px);
+    outline: 10px solid;
+  }
```
Cahllagerfeld, line 40: we removed this block, because this styling actually isn't for production. This blurs the image in case a `alt`-prop is missing. Its more for the development process that we dont forget this, but this rule shouldn't be used in production

Cahllagerfeld (commented):
What do you think about this?

src/lib/assets/global.css
nisarhassan12: Thanks, @Cahllagerfeld. I'll remove this as Svelte also gives an accessibility warning if an alt tag is missing.

Cahllagerfeld (approved):

PR #1513, opened 2022-01-20
refactor(gp-snapshots): Mention the cli option

axonasif:
## Description
There is a cli option for `gp` called `snapshot` but it is not reflected in the docs.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


mikenikles:
Thank you Asif. Do you mind updating https://www.gitpod.io/docs/command-line-interface as well please?

axonasif:
> Thank you Asif. Do you mind updating https://www.gitpod.io/docs/command-line-interface as well please?

Sure! Thank you for pointing out ❤️

axonasif:
@mikenikles In https://www.gitpod.io/docs/sharing-and-collaboration#sharing-snapshots
We have a line: `And of course, no access to any of your credentials is shared.`, is this correct then?
If not, let's change it fast because it seems to be conflicting with the security note at the end of the doc.

axonasif:
@mikenikles can you review again now 😄 ?

mikenikles (approved):

PR #1526, opened 2022-01-25
docs(prebuilds): clarify example about user variables

trumbitta:
## Description
I wasn't sure if user specific environment variables were available in `before` scripts (which are the suggested scripts where to globally install npm packages from), so I did my tests and found out.

I also found out that `gp` is an alias for `git pull` in the `git` plugin for `oh-my-zsh` so I had to specify the full path.
Given how popular `oh-my-zsh` is, and that the `git` plugin is enabled by default, I thought it'd be safer to also propose this change to the provided example.

## Release Notes

```release-note
updated example on how to use user specific environment variables in prebuilds
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1526"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1526"></a>

### DIFF
--- a/gitpod/docs/prebuilds.md
+++ b/gitpod/docs/prebuilds.md
@@ -209,12 +209,16 @@ If you don't want the comments to be added, disable them using `addComment: fals
 
 ## User specific environment variables in prebuilds
 
-It is not necessarily best practice to have user specific environment variables in a prebuild `init` block, but sometimes there are build time requirements that mean certain tokens need setting or files need creating. Environment variables defined within your Gitpod Variables preferences are not imported by default, but they can be accessed with the following command within an `init` block:
+It is not necessarily best practice to have user specific environment variables in a prebuild `before` or `init` block, but sometimes there are build time requirements that mean certain tokens need setting or files need creating. Environment variables defined within your Gitpod Variables preferences are not imported by default, but they can be accessed with the following command within a `before` or `init` block:
 
 ```yaml
 tasks:
+  - before: |
+      eval $(command gp env -e)
+      echo "Hello ${MY_VAR}"
   - init: |
-      eval $(gp env -e)
+      eval $(command gp env -e)
+      echo "Hello ${MY_VAR}"
 ```
 
 After that, the available environment variables will be installed into the rest of you shell script and can be accessed normally.


axonasif (changes requested):

gitpod/docs/prebuilds.md
```diff
@@ -209,12 +209,16 @@ If you don't want the comments to be added, disable them using `addComment: fals
...
+It is not necessarily best practice to have user specific environment variables in a prebuild `before` or `init` block, but sometimes there are build time requirements that mean certain tokens need setting or files need creating. Environment variables defined within your Gitpod Variables preferences are not imported by default, but they can be accessed with the following command within a `before` or `init` block:
 
 ```yaml
 tasks:
+  - before: |
+      eval $(/usr/bin/gp env -e)
+      echo "Hello ${MY_VAR}
   - init: |
-      eval $(gp env -e)
+      eval $(/usr/bin/gp env -e)
```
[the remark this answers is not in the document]
trumbitta: I chose `command` because it reminded me something I knew from the old times 🧓

axonasif (approved):
Thank you very much for contributing @trumbitta 🎉

gitpod/docs/prebuilds.md
```diff
@@ -209,12 +209,16 @@ If you don't want the comments to be added, disable them using `addComment: fals
...
 ## User specific environment variables in prebuilds
 
-It is not necessarily best practice to have user specific environment variables in a prebuild `init` block, but sometimes there are build time requirements that mean certain tokens need setting or files need creating. Environment variables defined within your Gitpod Variables preferences are not imported by default, but they can be accessed with the following command within an `init` block:
+It is not necessarily best practice to have user specific environment variables in a prebuild `before` or `init` block, but sometimes there are build time requirements that mean certain tokens need setting or files need creating. Environment variables defined within your Gitpod Variables preferences are not imported by default, but they can be accessed with the following command within a `before` or `init` block:
 
 ```yaml
 tasks:
+  - before: |
+      eval $(command gp env -e)
+      echo "Hello ${MY_VAR}
```
axonasif, line 218: It's missing a closing double-quote `"`
Also I think that putting an example here might not be on point because the user is looking for this as they already find that some env var is empty or not working.

axonasif (changes requested):

gitpod/docs/prebuilds.md
trumbitta: Oof copy-pasting from my terminal on lunch break with a ham sandwich being wolfed down. It's fixed now 🙏 

About the example in there, I just thing obvious always wins. It's for giving context and to anticipate and reinforce the closing sentence: "After that, the available environment variables will be installed into the rest of you shell script and can be accessed normally."

You know, like the same thing having a picture near paragraph of text does.

But if I didn't convince you, just say the word and I'll delete it.

axonasif: Fair enough, it's just something that I had thought 😄

axonasif (approved):

shaal:
@trumbitta @axonasif AFAIK, User environment variables that you want to use during prebuild, are only available in prebuild that were triggered manually.
There are prebuilds that happen automatically (new commits, Gitpod update, etc) - that will run Prebuild and it will not have access to your user environment variables.

trumbitta:
> @trumbitta @axonasif AFAIK, User environment variables that you want to use during prebuild, are only available in prebuild that were triggered manually. There are prebuilds that happen automatically (new commits, Gitpod update, etc) - that will run Prebuild and it will not have access to your user environment variables.

@shaal I worked on this today, and I have prebuilds accessing my vars like this that are generated by commits on a MR in GitLab

trumbitta:
> I know this because I added IFTTT hook, to send me an email every time a prebuild runs. https://github.com/shaal/DrupalPod/blob/main/.gitpod.yml#L7
> 
> And I can see that whenever it's something I generated (commit I made) - env vars are loaded, but whenever it's generated by someone else commits, or Gitpod update - env vars are not accessible during prebuilds.

Ok this is already slightly different from what you said above, and something I didn't try :) but this PR is just a refactor of instructions that were already there. Do you think a different PR is worth it? What you are describing sounds more like a bug.

shaal:
@trumbitta it might be a bug? 
Currently, users without write access to a repo, can trigger manual prebuild, and when depending on env vars, it can become an issue.
I didn't test it yet, but it's possible that **_project_** env vars would solve this problem.

trumbitta:
❤️

PR #1534, opened 2022-01-26
Remove vite from package.json

mishrasatyam:
## Description

Since `vite` is installed when `svelte-kit` is installed, we can remove it as direct dependency.
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1534"></a>


Cahllagerfeld:
Thanks for your contribution 🚀 I think it looks good 👍  I'd like to get the opinion from @mikenikles before merging hastily though.

mikenikles:
It looks like that was introduced when we migrated from Sapper to SvelteKit and may have been a required dependency back then ([diff](https://github.com/gitpod-io/website/commit/dbd48199106bd1f681810e07a2a97c5fc527116b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R82)).

We've come a long way, thank you @mishrasatyam for the PR 🙏.

PR #1551, opened 2022-01-28
Create Button / Link-Button Component

Cahllagerfeld:


src/routes/blog/github-codespaces.md
```diff
@@ -12,6 +12,10 @@ title: GitHub Codespaces, Welcome to the Party!
   export const prerender = true;
 </script>
 
+<script>
+  import LinkButton from "$lib/components/ui-library/link-button/link-button.svelte"
```
Cahllagerfeld, line 16: I need to use the import of the `.svelte` file here, otherwise I get the Error that `<LinkButton> is not a SSR-Component`. I did some research and it seems to be related to the `generate: ssr` in the `svelte.config`. What do you think about this @mikenikles?

Cahllagerfeld (commented):
I'm still not sure, if the `d.ts` files are worth to keep...

PR #1560, opened 2022-02-02
Update landing page with JetBrains screenshots.

### LINKED ISSUE #1327: Update landing page with JetBrains screenshots
The landing page needs to be updated with screenshots for all JetBrains IDEs that are currently in beta
Bildschirmfoto 2021-12-06 um 13 37 11

**To do's:**

- [ ] take screenshots (size: 1250 x 800, zoomed-in, similar style to the existing screenshots, I used the Go examples) @loujaybee [[1](https://github.com/gitpod-io/website/pull/1225#issuecomment-966271470)]
- [ ] adjust screenshots with the current design @ChristinFrohne 
- [x] switch the toggle to Desktop when selecting any Jetbrains IDE @nisarhassan12

nisarhassan12:
Fixes #1327

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1560"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1560"></a>


src/lib/components/index/section-ides.svelte
```diff
@@ -6,13 +6,25 @@
...
   }
 
+  $: if (
+    ides.find(({ name }) => name === selectedIde).screenshots.browser ===
+    undefined
+  ) {
+    toggleChecked = true;
+    disabled = true;
+  } else {
+    toggleChecked = false;
```
mikenikles, line 24: @nisarhassan12 This fixes the issue you shared internally when you select a JetBrains IDE and go back to VS Code and had to click twice on the toggle.

The problem was that the hidden input field's `checked` state wasn't in sync with the `toggleChecked` variable.

ChristinFrohne (commented):
Awesome, thanks @nisarhassan12  🎉

a few comments:

- [ ] could you swap the GoLand screenshot with the version [here](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=514%3A2770)? Currently, the alignment of the browser bar is a bit off. Sorry, my bad!
- [ ] for the toggle: I don't think we need to grey it out. We can keep it as it is, but automatically select "Desktop" and disable the possibility to switch to "Browser".

nisarhassan12:
> for the toggle: I don't think we need to grey it out. We can keep it as it is, but automatically select "Desktop" and disable the possibility to switch to "Browser".

Thanks, @ChristinFrohne I personally think we should have a visual disabled state for the toggle here just like we have it for the Switch and the button components in our Component library.

When the users are switching to a different ide via hovering on the IDEs logos on the right they will try to click on the toggle to see their browser versions and nothing will happen with the toggle they will eventually figure it out but not right away.

WDYT? your call here? I'm happy :) to leave out the current greying of the toggle for the disabled state. Thanks

src/lib/components/index/section-ides.svelte
nisarhassan12: Thanks, @mikenikles for fixing this.

phimae:
Sorry if I'm jumping in here, was checking which open PRs we have before Nisar goes on vacation. As he is working late today I'll try to unblock:

> they will try to click on the toggle to see their browser versions and nothing will happen with the toggle they will eventually figure it out but not right away.

I agree, better to grey out imo @nisarhassan12

nisarhassan12:
Thanks, @phimae for jumping here. Sorry that I didn't mention it here @ChristinFrohne brought this up internally on Slack [here](https://gitpod.slack.com/archives/C020VCB0U5A/p1644486024620839) we decided on what George suggested and I have just updated the PR with that. 

Chirstin please re-review the PR when you get time. Thanks

ChristinFrohne:
@nisarhassan12 could you do a follow-up PR once you are back to reduce the corner radius on mobile? Thanks!
Bildschirmfoto 2022-02-11 um 08 48 39

PR #1569, opened 2022-02-07
Update Docs Layout

### LINKED ISSUE #1532: Design clean-up for  /docs
**Problem**
The documentation page is a bit messy and needs a clean-up and a better alignment. 

|status quo| suggestion (very rough mockup)|
|-|-|
|Bildschirmfoto 2022-01-26 um 08 40 54|Bildschirmfoto 2022-01-26 um 08 58 20

Actions:

- [x] align the search bar size with the width of the text
- [x] align the top of the secondary navigation with the navigation on the left
- [x] color "on this page" in black (#12100C) as all headlines are black
- [x] move the "edit in Gitpod" button next to the search bar and align it with the secondary navigation
- [x] add little arrows to submenus similar to the way [Tailwind](https://tailwindcss.com/docs/upgrade-guide) does it

We might need to reduce the width of the text block to fit the secondary navigation nicely. See [Tailwind](https://tailwindcss.com/docs/upgrade-guide) for inspiration

Cahllagerfeld:
closes #1532

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1569"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1569"></a>


Cahllagerfeld (commented):
The Mockup suggests removing the Search-Shortcut from the Placeholder of the input-field. Do we  want to do this? My concern is that noone knows this shortcut..

mikenikles:
It's in [this `if`](https://github.com/gitpod-io/website/blob/main/svelte.config.js#L76) statement where we can modify the ToC node.

mikenikles:
> The Mockup suggests removing the Search-Shortcut from the Placeholder of the input-field. Do we want to do this? My concern is that noone knows this shortcut..

I share this concern. @ChristinFrohne, what do you think? Your call 📞.

mikenikles (approved):
Code 👍. Let's wait for Christin's decision on the keyboard shortcut, but otherwise let's get this merged.

ChristinFrohne:
>The Mockup suggests removing the Search-Shortcut from the Placeholder of the input-field. Do we want to do this? My concern is that noone knows this shortcut..

sorry about the confusion. That change was not indented, let's keep the shortcut :) 

Thanks a lot for addressing this issue 🙏 🧡  
Here are a few more things I noticed:

- [ ] the shadow around the boxes is cut
Bildschirmfoto 2022-02-09 um 08 15 22

- [ ] the style of the search bar doesn't align with the component in Figma. E.g. the width should be 48px in alignment with the buttons and its background color is #F9F9F9 [[1](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=514%3A2831)]


- [ ] instead of using ">" for sub-topics. we have this icon we should use. It is the same one as for the FAQ section, just rotated and scaled.
fold

- [ ] what do you think above using 16px font size for the navigation? By that, we create clearer hierarchies and it becomes easier for the user to focus on one thing. e.g. Tailwind uses font sizes similarly

Cahllagerfeld (commented):
I managed to get this finally working the the TOC and the arrow-image. The only thing that is currently left, is the grouping of the IDEs in the left-side menu. I'm not quite sure if we are able to implement this thing with the Menu how it currently is @mikenikles, without refactoring or even rewriting the whole left-side menu. I think we already spoke about some a11y things for the left-side menu. Do you think its worth doing this rewrite now, or wait until we tackle the a11y?

PR #1594, opened 2022-02-10
/for/startups updates.

### LINKED ISSUE #1585: Include case study and new quote in /for/startups
Hi there! Now that we have case studies, we should inject them in our existing material. For the `/for/startups` page we should do the following:

- [x] Swap the existing quote against the new quote box style (can be the same box as on the /customers page)
image

- [x] Add a new section below "Accelerate your workflow" that looks something like this (@ChristinFrohne will provide a more thoughtful mockup) If the page becomes too crowded, we can drop the "Code from anywhere on any device" section 
image

nisarhassan12:
Fixes #1585 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1594"></a>


phimae (approved):
lgtm

PR #1607, opened 2022-02-15
Minor copyedit (grammar and spelling).

metcalfc:


Cahllagerfeld:
I just approved the Netlify deployment manually. Whats a bit strange is, that the deployment should automatically run for all inside our Github-Org (afaik though). Do we need to adjust something here @mikenikles

PR #1632, opened 2022-02-18
feat: Use lightweight workspace image and globally set HMR_HOST

axonasif:
## Description

- We can use the new lightweight `workspace-node` image to speed up the workspace loading time.
- `HMR_HOST` env var was only scoped to the `.gitpod.yml` `command:` task-terminal, but now all terminals will inherit it.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1632"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1632"></a>


.gitpod.Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM gitpod/workspace-node:latest
+
+RUN printf '%s\n' 'export HMR_HOST="$(gp url 3000)"' >> $HOME/.bashrc
```
mikenikles, line 3: I recently thought about an even simpler approach. We don't need this env variable and can leverage `GITPOD_WORKSPACE_URL` instead, which is set on all workspaces.

In `svelte.config.js`, use the following:

```js
vite: {
      server: {
        hmr: {
          clientPort: process.env.GITPOD_WORKSPACE_URL ? 443 : 3000,
          host: process.env.GITPOD_WORKSPACE_URL
            ? process.env.GITPOD_WORKSPACE_URL.replace("https://", "3000-")
            : "localhost",
        },
      },
```

axonasif: That's a really good approach 🙏

PR #1649, opened 2022-02-21
fix: alignment of chagenlog buttons

Cahllagerfeld:


gtsiolis:
Thanks for the quick fix, @Cahllagerfeld! 🏀

PR #1650, opened 2022-02-21
Include Site Visitor Cookie on Segment Load

jakobhero:


jakobhero:
woops, just saw that @mikenikles already addresses the expiry of `gitpod-marketing-website-visited` which caused the need for this PR  in #1640, closing this!

PR #1662, opened 2022-02-23
Add Tailwind-Prettier-Plugin

Cahllagerfeld:


Cahllagerfeld (commented):
This will cause quite a few merge-conflicts. We should wait with Merging until the other High-Prio PRs are merged

Cahllagerfeld:
Due to all Merge Conflicts we should close this, and run prettier against the current state of the repo whenever we need it. The Installation of the Prettier-Tailwind-Plugin isn't hard and when keeping it open we have to maintain this PR, which produces overhead in my opinion.

PR #1707, opened 2022-03-03
Segment for Email-Widget

Cahllagerfeld:
Adds Segment Support for the new E-Mail Widget

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1707"></a>


src/lib/components/blog/e-mail-widget.svelte
```diff
@@ -18,11 +18,11 @@
...
-    // trackEvent("blog_email_submitted", {
-    //   email: email,
-    //   url: window.location.href,
-    //   path: window.location.pathname,
-    // });
+    trackIdentity({
+      email_untrusted: email,
+    });
+
+    trackEvent("email_submitted", {});
```
jakobhero, line 25: i shortened the name of the event a little bit so that it's easier to work with in the background. also, for a consistent event naming, we avoid hyphens in favour of underscores.

jakobhero (approved):
LGTM

src/lib/components/blog/e-mail-widget.svelte
Cahllagerfeld: Thanks for fixing it 🚀

PR #1711, opened 2022-03-03
Layout of Security Cards

Cahllagerfeld:
- [x] Spacing inside cards
- [x] spacing between cards
- [x] padding of SVG-Files

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1711"></a>


Cahllagerfeld:
Hi @ChristinFrohne, I refactored the whole Security-Program thing to use the exact same Card-Component, we use on the index-page. Other than this I enhanced the Card-Component, so we can pass in if the text should be left-center, or right-aligned

ChristinFrohne (approved):
Awesome, thanks <3

Issue #1182, opened 2021-10-29
Implement an end-to-end testing framework based on https://www.cypress.io/


mikenikles:

mikenikles:
As a first step, we should review https://www.cypress.io/blog/2021/05/24/full-testing-of-html-emails-using-ethereal-accounts/ and implement end-to-end testing for all forms on the website.

mikenikles:
End-to-end tests should probably also run as a GitHub action for each PR, ideally against the Netlify PR deployment URL.

mikenikles:
Also consider to validate links that point to anchors and make sure these anchors exist. This is frequently used to link between documentation pages. ([internal discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1657694360026829?thread_ts=1657638470.784789&cid=C01KGM9ETU6))

PR #1722, opened 2022-03-07
e2e for the forms.

nisarhassan12:
Addresses parts of #1182 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1722"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1722"></a>


.gitpod.Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM gitpod/workspace-node:latest
```
nisarhassan12, line 1: @mikenikles I didn't knew/was not able to figure out how to use multiple Docker images within a `.gitpod.yml` so moved it here from the config file. Does it look good?

nisarhassan12: Also, it took quite a long time for the workspace to start and the pre-build to finish maybe it's the 2 FROM that are having an impact here?

cypress/integration/sales_spec.js
```diff
@@ -0,0 +1,15 @@
+describe("Test the sales contact form", () => {
+  it("Fills out the form data", () => {
+    cy.visit("localhost:3000/contact/sales");
```
nisarhassan12, line 3: @mikenikles Currently the specs work locally. I don't know How can I get the URL for a workspace to be used here. Any suggestions? Thanks

.gitpod.Dockerfile
mikenikles: We can remove this line. The `gitpod/workspace-full-vnc` image we use below is based on `gitpod/workspace-full` which contains Node.js [[1](https://github.com/gitpod-io/workspace-images/blob/main/legacy/full/Dockerfile#L120)]

cypress/integration/sales_spec.js
mikenikles: Have a look at the `baseUrl` configuration (see [docs](https://docs.cypress.io/guides/references/configuration#Global))

cypress/integration/sales_spec.js
```diff
@@ -0,0 +1,15 @@
...
+    cy.wait(100);
```
mikenikles, line 4: We shouldn't use `cy.wait(time)`. Please read the docs [[1](https://docs.cypress.io/api/commands/wait#Time)] and follow the best practices linked there.

cypress/integration/sales_spec.js
```diff
@@ -0,0 +1,15 @@
...
+    cy.get('[data-test-id="Self-hosting"]').click();
```
mikenikles, line 5: What do you think about using https://testing-library.com/docs/cypress-testing-library/intro/ instead of `data-test-*` attributes?

Please [see this chapter](https://testing-library.com/docs/) for an introduction to the testing library.

If you look at this PR, you had to manually add `data-test-*` attributes. With the testing library, we query for elements the way users query for elements, i.e. by looking for buttons with certain text.

package.json
```diff
@@ -26,7 +26,8 @@
     "start": "svelte-kit preview",
     "validate": "svelte-check",
     "prepare": "husky install",
-    "generate:blog-post": "plop --plopfile ./generators/index.cjs blog-post"
+    "generate:blog-post": "plop --plopfile ./generators/index.cjs blog-post",
+    "cypress:open": "cypress open --browser firefox"
```
mikenikles, line 30: Leaving this here as a reminder: We'll also need a `test:e2e` script that runs Cypress in the command line.

mikenikles (changes requested):

cypress/integration/sales_spec.js
nisarhassan12: @mikenikles I had a look at the docs Best Practices and the following is what it says:

image

but when I run the spec without adding `cy.wait(100)` it fails to select the radio button label with a click. I have no idea why that might be see the below video:

https://user-images.githubusercontent.com/46004116/157433028-42847d72-d251-4f57-8036-0bc4fd501ac0.mov

If I do it with `cy.wait(100)` then the 'Self-hosting' option is selected as expected see the below video:

https://user-images.githubusercontent.com/46004116/157434061-e8012950-db8b-4369-b163-044c275dc089.mov

cypress/integration/sales_spec.js
nisarhassan12: Thanks, Mike for sharing this I didn't know about this. I just migrated the existing specs to use it + also wrote the spec for the [Enterprise License](https://www.gitpod.io/enterprise-license) page using it.

cypress/integration/support_spec.js
```diff
@@ -0,0 +1,14 @@
...
+  it("Fills out the form data", () => {
+    cy.visit("/contact/support");
+    cy.wait(100);
+    cy.findByText(/billing/i).click();
+    cy.findByLabelText(/name/i).click().type("Gitpod User");
+    cy.findByLabelText(/mail/i).click().type("gitpoduser@gitpod.io");
+    cy.findByLabelText(/message/i)
+      .click()
+      .type("Gitpod is Awesome!");
+    cy.findByTestId("consent").click();
```
nisarhassan12, line 11: @mikenikles I have gotten rid of all the other `data-test-id` attributes apart from the ones used for the checkboxes as I was not able to figure out how to target them with the testing library.

nisarhassan12, line 11: @mikenikles I have gotten rid of all the other `data-test-id` attributes apart from the ones used for the checkboxes as I was not able to figure out how to target them with the testing library.

nisarhassan12, line 11: @mikenikles I have gotten rid of all the other `data-test-id` attributes apart from the ones used for the checkboxes as I was not able to figure out how to target them with the testing library.

package.json
nisarhassan12: Thanks, Mike I did that in https://github.com/gitpod-io/website/pull/1722/commits/37455beb136c01664c4b0375d949ef68450dbead.

nisarhassan12:
@mikenikles Also not sure why but today the specs seem to be failing in Gitpod while they work reliably locally. The following is what I'm getting:
Screenshot 2022-03-10 at 15-40-32 gitpodio-website-ffbg6o8sh14 0 - noVNC

cypress/integration/sales_spec.js
mikenikles: Some race condition or what not that happens there. I couldn't find a solution without refactoring the form input listeners (`on:change` and `bind:checked`).

Let's keep the wait in there as it's an isolated case with no quick fix.

mikenikles (approved):
That's a good starting point. It doesn't fix #1182 yet though (see [this comment](https://github.com/gitpod-io/website/issues/1182#issuecomment-1047919611)).

The tests in this PR don't actually test that the email is sent and received, neither do the tests validate the email content is correct.

In the interest of unblocking this, let's merge but keep #1182 open.

PR #1771, opened 2022-03-14
add the skip nav link.

### LINKED ISSUE #1766: [accessibility]: add skip links
Currently, the site is missing skip links which means that the users who can't use a mouse due to some disability or by choice via a keyboard or switch will have to tab through all the nav links each and every time they visit a page on our site which is not good from an accessibility point of view. 

Sites usually have skip links in place to avoid the unnecessary round trips of the navigation. The below screenshot shows examples of How GitHub and NYT do it.

image

image

http://web-accessibility.carnegiemuseums.org/code/skip-link/

nisarhassan12:
Fixes #1766 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1771"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1771"></a>


ChristinFrohne (commented):
- [x] Could you add some spacing to the sides?

see vercel vs ours
Bildschirmfoto 2022-03-15 um 09 00 44
Bildschirmfoto 2022-03-15 um 09 00 39


- [ ] in comparison to Vercel, you need one extra button click to move to the navigation menu. Would be great if you can just align the style with Vercel

Otherwise it looks good to me

nisarhassan12:
> - [ ] in comparison to Vercel, you need one extra button click to move to the navigation menu. Would be great if you can just align the style with Vercel

@ChristinFrohne did you mean the following behaviour? notice how it happens for the first time on the home page that pressing tab and then enter does not seem to take to the main area but afterwards, on the other pages, it seems to work as expected or did you mean something else? Thanks

https://user-images.githubusercontent.com/46004116/158353492-5862d90c-955c-4c4a-ab87-65c05c2bdbcf.mov

nisarhassan12:
@mikenikles I have no idea why it is the case. Can you have a look when you get some time? Thanks

nisarhassan12:
@Cahllagerfeld Please have a look when you get time. Please specifically look at the following comment from Christin 

> - [ ]  in comparison to Vercel, you need one extra button click to move to the navigation menu. Would be great if you can just align the style with Vercel

I'm not able to fully understand or reliably reproduce this issue.

Cahllagerfeld (commented):
For me, I couldn't test it in Firefox - when hitting Tab there my page begins to scroll, instead of tabbing though it. On Brave the Dialog appeared, but as Christin mentioned, when this popup is shown, I need two key presses to get to the nav-items

nisarhassan12:
> For me, I couldn't test it in Firefox

Same for me for testing on Firefox on a Mac you would have to turn on the Voice Over via `CMD + F5`.

> I need two keypresses to get to the nav-items

For some reason I'm unable to reproduce what both of you are experiencing for me it works as expected i.e you tab once the "Skip to Content" link appears and the next time you tab it takes you to the first link in the nav which is the Gitpod logo pointing to the Homepage. 

@mikenikles Can you reproduce what Julain and Christin encountered? Thanks

nisarhassan12:
> I need two key presses to get to the nav-items

Mike and I looked at it in the call and it turns out what both of you are referring to is expected behaviour the reason is that the first tab is for the Gitpod logo which takes you to the home page and then the rest of the Nav items start with the second Tab key press.

PR #1775, opened 2022-03-17
Add Canonical URL

Cahllagerfeld:


Cahllagerfeld (commented):
Hi @nisarhassan12 I implemented the canonical url thing as I discussed it with Mike yesterday. This may be a solution when having an instance of the website deployed at vercel and at netlify, to always point to `https://www.gitpod.io`. I ignored the trailing slash at the end currently. What do you think about this?

PR #1806, opened 2022-03-23
 [Bug]: Empty lines with escaped periods.

### LINKED ISSUE #1659: [Bug]: Empty lines with escaped periods
### Page URL

https://www.gitpod.io/docs/ides-and-editors/vscode

### User feedback

When the viewport is between 1391 and 1427 pixels wide, you can observe the period in the `Managing Extensions` section being escaped to its own line, which looks... not so good.
image

I am sure this is not the only page it is happening on but it struck me since its the width my browser is on my Macbook.

nisarhassan12:
fixes #1659


Cahllagerfeld:
Thats actually funny, because I stumbled across the same issue earlier today :D I thought it relates to this discussion with Mike and Christin where we spoke about the `nbsp` for headlines

PR #1836, opened 2022-03-29
Beam users directly into the product (/workspaces)

svenefftinge:
## Description
This PR beams users directly into the product `https://gitpod.io/` instead of the intermediate step of moving to `https://www.gitpod.io/#getting-started`. It also changes the links in the getting started section to point to `/https://gitpod.io/` instead of `https://gitpod.io/new` which is too involving for starters.

## Reasoning behind this experiment

We offer new users the same choice in the product. First users pick the git provider, before they land in workspace and can pick one of their own or a template repo to start their first workspace.

## Related Issue(s)
Fixes https://github.com/gitpod-io/gitpod/issues/8988

## How to test
Go to preview website and sign up without choosing a template.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1836"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1836"></a>


jldec:
I think this change would simplify the onboarding flow and I'm generally in favor of fewer steps.

1. **(this needs to be addressed before merging)** We need to be careful with un-authed users since they are currently redirected from https://www.gitpod.io/ back to www.gitpod.io. All the links in https://www.gitpod.io/#get-started point to a path in the dashboard which does not redirect back for un-authed users. To repro the problem, open the [preview for this PR](https://deploy-preview-1836--gitpod-kumquat.netlify.app/) in an incognito browser window.

2. (this is more of an FYI) The list of example templates exists in the docs and a subset in https://www.gitpod.io/#get-started but we no longer show that list in the new workspace popup. This reduces the chance that users will discover the examples.

svenefftinge:
> (this is more of an FYI) The list of example templates exists in the docs and a subset in https://www.gitpod.io/#get-started but we no longer show that list in the new workspace popup. This reduces the chance that users will discover the examples.

What was the reasoning behind showing a different set of examples in the product?

src/lib/components/index/get-started/link-git-repository.svelte
```diff
@@ -73,7 +73,7 @@
     Continue with GitLab
   </LinkButton>
   <LinkButton
-    href="https://gitpod.io/new"
+    href="https://gitpod.io"
```
phimae, line 76: should also point to `/workspaces` I guess

src/lib/components/index/get-started/link-git-repository.svelte
```diff
@@ -92,7 +92,7 @@
   <LinkButton
     size="large"
     variant="unstyled"
-    href="https://gitpod.io/new"
+    href="https://gitpod.io"
```
phimae, line 95: should also point to `/workspaces `

src/lib/contents/github-page.ts
```diff
@@ -53,7 +53,7 @@ export const ghTeacherPricingPlans: Pricing[] = [
       "30min Timeout",
     ],
     btnText: "Try Now",
-    btnHref: "/#get-started",
+    btnHref: "https://gitpod.io/",
```
phimae, line 56: should also point to `/workspaces `

src/lib/contents/home/features.ts
```diff
@@ -85,7 +85,7 @@ export const otherFeatures: Feature[] = [
   {
     ...codeAnywhereFeature,
     moreButton: {
-      href: "#get-started",
+      href: "https://gitpod.io/",
```
phimae, line 88: should also point to `/workspaces `

src/lib/contents/pricing.ts
```diff
@@ -113,7 +113,7 @@ export const empowermentFeatures: Card[] = [
 
 export const freePlanTableData: FeatureTableColumn = {
   link: {
-    href: "/#get-started",
+    href: "https://gitpod.io/",
```
phimae, line 116: should also point to `/workspaces `

phimae (commented):
I think @jldec is right, there are a few instances (I added comments in the code where) where `/new` got replaced with `/` and not with `/workspaces` so people wouldn't be able to sign up I guess

**In general:**
Replacing `/new` makes a ton of sense given lower retention numbers we've seen for people that create a project right away.
Let's closely monitor the impact on signups with a template repo as first context since we're making this choice a little less obvious now. Maybe the cmd+O modal still suffices for template selection.

PR #1848, opened 2022-03-31
[Analytics] Pass Signup Context implicitly

jakobhero:
Determines signup context in Segment Calls implicitly. This improves the status quo as it no longer requires that the data-analytics object is manually populated when a new signup button is implemented.
A context is classified as signup if a user clicks on a link that redirects to the dashboard and the `gitpod-user` cookie is not set to true.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1848"></a>


Cahllagerfeld (approved):
Looks good to me 🛹 👍

phimae (approved):
👍 works, checked in the staging debugger in segment

analytics.track('website_clicked', {
  authenticated: false,
  context: 'signup',
  destination: 'https://gitpod.io/workspaces/',
  label: 'Try Now',
  path: '/',
  position: 'hero',
  url: 'https://deploy-preview-1848--gitpod-kumquat.netlify.app/',
  variant: 'link'
});

PR #1855, opened 2022-04-03
[Analytics] Remove redundant `product_cta_clicked` track call

jakobhero:
Removes `product_cta_clicked` track call. The call is made redundant through `website_clicked`, which contains information about which button was clickced in the properties.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1855"></a>

### DIFF
--- a/src/lib/components/jetbrains-space/hero.svelte
+++ b/src/lib/components/jetbrains-space/hero.svelte
@@ -37,12 +37,7 @@
       class="mt-x-small"
       variant="primary"
       size="large"
-      href="https://gitpod.io/workspaces/"
-      on:click={() =>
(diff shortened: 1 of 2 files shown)


Cahllagerfeld (commented):
Is it correct, that both trackers are getting removed, or should only one be removed? 😅 🤔

Cahllagerfeld:
just a small ping @jakobhero. If my comment above is irrelevant we're good to go I think 👍

jakobhero:
thanks for the ping @Cahllagerfeld! it's correct that both of them should be removed. the event is made redundant through the collection of a different event, `website_clicked`, so we don't need any event calls of `product_cta_clicked` on the website anymore!

PR #1860, opened 2022-04-05
[Test] compress images manually

Cahllagerfeld:


Cahllagerfeld (commented):
I run all images from the lighthouse report through [https://tinypng.com/](https://tinypng.com/), just out of curiosity to see if it has an impact

Cahllagerfeld:
Seems like its not making a huge difference

Cahllagerfeld:
It looks like its not making a huge differnce compressing them manually

PR #1878, opened 2022-04-07
implement lazy loading.

### LINKED ISSUE #1850: Lazy load images to enhance page load times.
On the [Changelog](https://www.gitpod.io/changelog) the lighthouse score is 50 and that is because of the fact that all the changelog entries images are loaded on the initial page load.

image

The problem with loading all the images that a page has is that we are never sure if for instance a user will scroll down and see them all thus resulting in unnecessary usage of bandwidth which we can avoid via lazy loading. Lazy loading will help improve the overall performance scores as well.

nisarhassan12:
Fixes #1850 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1878"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1878"></a>


mikenikles:
@nisarhassan12 What's the status of this?

PR #1914, opened 2022-04-14
Add Settings Sync Docs

filiptronicek:
## Description

This adds mainly the general idea behind why you cannot sync to Gitpod without this and the guide on how to enable the Settings Sync itself. It does not include anything about what it even is, which may be a good thing to add.

This also fixed syntax highlighting for a code block on the VS Code Desktop page. 

Corresponding issues and PRs: https://github.com/gitpod-io/openvscode-server/pull/337 https://github.com/gitpod-io/gitpod/issues/9136

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1914"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1914"></a>


akosyakov:
It is not related to VS Code Desktop though only. I think it is worth to have a dedicated VS Code Setting Sync page and explain how settings can be synced between all installation of VS Code (Web and Desktop) in context of Gitpod.

A general introduction why we need it and how it works would be nice too.

filiptronicek:
Agreed, having its own page sounds like the way to go.

gitpod/docs/ides-and-editors/settings-sync.md
```diff
@@ -0,0 +1,49 @@
...
+title: VS Code Settings Sync
+---
+
+<script context="module">
+  export const prerender = true;
+</script>
+
+# {title}
+
+VS Code is so beloved hugely because of its endless possibilities for customization. A problem arises when there are multiple machines or VS Code instances you'd like to have with the same extensions, themes and settings.
```
akosyakov, line 12: I think it makes sense here to reference that each workspace start is a new machine basically. Maybe link to https://www.gitpod.io/docs/workspaces for the context about ephemeral nature of Gitpod

akosyakov:
It looks good to me ❤️ We should deploy an extension first to merge. 

@loujaybee Could you check whether it makes sense to you now?

gitpod/docs/ides-and-editors/settings-sync.md
```diff
@@ -0,0 +1,49 @@
...
+
+The solution for exactly this and a bunch more is called Settings Sync. It saves all your preferences to the cloud so that you simply log in to the same account on all of your instances and have all your settings in sync.
+
+## How does it work?
+
+When Settings Sync is running, it polls the backend for changes on an interval and at the same time makes sure that after any change to your preferences, these changes are pushed back. After the initial (first) sync, only the changed items are pushed to the backend server to save bandwidth and make the syncronization nearly instant.
+
+The following items are synced when you use Settings Sync:
+
+- Settings - your VS Code preferences [[docs](https://code.visualstudio.com/docs/getstarted/settings)]
```
akosyakov, line 22: You are forgot about extensions! They are actually most tricky. Because users don't realize that they install extensions for all workspaces by default. And tehy should select `install (no sync)` to install only for given workspace. Maybe it should be mentioned somewhere.

filiptronicek: Fixed now.

gitpod/docs/ides-and-editors/settings-sync.md
```diff
@@ -0,0 +1,60 @@
...
+We don't have support for these environments, yet, but are actively looking into it.
+
+## Locally
+
+By default, your local VS Code installation is _most likely_ equipped with a pre-configured setup to sync all of your IDE-related settings via a GitHub or a Microsoft account. This means that your preferences are synced to Microsoft's servers and that means Gitpod has no access to them.
+
+You can change this fact and sync all of your preferences to Gitpod instead. You will not lose any data by doing this, but if you still want to sync your settings across machines we recommend you to do the same on all of them as well.
+
+1. Make sure the [Gitpod extension](https://marketplace.visualstudio.com/items?itemName=gitpod.gitpod-desktop) is installed and enabled.
+2. Hit <kbd>F1</kbd> to open the Command Pallete and run `Gitpod: Turn on Settings Sync`.
```
jeanp413, line 56: I renamed it to `Settings Sync: Enable signing in with Gitpod`
```suggestion
2. Hit <kbd>F1</kbd> to open the Command Pallete and run `Settings Sync: Enable signing in with Gitpod`.
```

gitpod/docs/ides-and-editors/settings-sync.md
```diff
@@ -0,0 +1,60 @@
...
+
+<script context="module">
+  export const prerender = true;
+</script>
+
+# {title}
+
+VS Code is so beloved hugely because of its endless possibilities for customization. A problem arises when there are multiple machines or VS Code instances you'd like to have with the same extensions, themes and settings.
+
+Every Gitpod workspace is basically its own machine, meaning that if you open a new workspace or restart an existing one, your unsynced preferences will not carry over ([why?](https://www.gitpod.io/docs/workspaces)). The solution for exactly this and a bunch more is called Settings Sync. It saves all your preferences to the cloud so that you simply log in to the same account on all of your instances and have all your settings in sync.
```
akosyakov, line 14: [Settings Sync](https://code.visualstudio.com/docs/editor/settings-sync) allows you to share VS Code settings and extensions between Gitpod and other instances of VS Code. Each start of [Gitpod workspace](https://www.gitpod.io/docs/workspaces) is its own instance of VS Code.

akosyakov:
I think we should add a reference to this doc from VS Code Browser and VS Code Desktop pages? Some short reference that it allows to share settings and extensions between Gitpod workspaces and other instance of VS Code and a link to a new page?

gitpod/docs/ides-and-editors/settings-sync.md
filiptronicek: Coolsies, thanks!

gitpod/docs/ides-and-editors/settings-sync.md
```diff
@@ -0,0 +1,60 @@
...
+
+## How does it work?
+
+When Settings Sync is running, it polls the backend for changes on an interval and at the same time makes sure that after any change to your preferences, these changes are pushed back. After the initial (first) sync, only the changed items are pushed to the backend server to save bandwidth and make the syncronization nearly instant.
```
andreafalzetti, line 18:
```suggestion
Once enabled enable Settings Sync it polls the backend for changes on an interval and at the same time makes sure that after any change to your preferences, these changes are pushed back. After the initial (first) sync, only the changed items are pushed to the backend server to save bandwidth and make the syncronization nearly instant.
```

Just a suggestion, feel free to drop it - but I started wondering: "What does it mean `running`? Is this something I can opt-in or it just runs as part of Gitpod?

andreafalzetti: `syncronization` -> `synchronization`

gitpod/docs/ides-and-editors/settings-sync.md
```diff
@@ -0,0 +1,60 @@
...
+We don't have support for these environments, yet, but are actively looking into it.
+
+## Locally
+
+By default, your local VS Code installation is _most likely_ equipped with a pre-configured setup to sync all of your IDE-related settings via a GitHub or a Microsoft account. This means that your preferences are synced to Microsoft's servers and that means Gitpod has no access to them.
+
+You can change this fact and sync all of your preferences to Gitpod instead. You will not lose any data by doing this, but if you still want to sync your settings across machines we recommend you to do the same on all of them as well.
+
+1. Make sure the [Gitpod extension](https://marketplace.visualstudio.com/items?itemName=gitpod.gitpod-desktop) is installed and enabled.
+2. Hit <kbd>F1</kbd> to open the Command Pallete and run `Settings Sync: Enable signing in with Gitpod`.
```
andreafalzetti, line 56:
```suggestion
2. Hit <kbd>F1</kbd> to open the Command Palette and run `Settings Sync: Enable signing in with Gitpod`.
```

andreafalzetti:
If I understand the docs correctly, the settings are synchronized server-side, so I would assume that while using Gitpod, regardless of me using `VS Code browser` or `VS Code Desktop`, I will have my settings synced?

gitpod/docs/ides-and-editors/settings-sync.md
filiptronicek: Thanks for pointing this out, saying that Settings Sync is "running" can be a bit misleading users, as it is just something they enable and don't have to check that it is "running". Settings Sync is a part of VS Code's source code and on Gitpod we just **enable** it automatically.

akosyakov:
> @andreafalzetti the settings sync is done from the VS Code client itself, with it running on the same level as extensions (I think 😄). This means we have to both authenticate the client (either Browser or Desktop) with Gitpod and then VS Code can start sending events and payloads. This means that there has to always be an extension that contributes a way to authenticate via Gitpod and which then gives VS Code a token for authenticating the requests to the Settings Sync endpoint on the backend.

> This simply means you can have Settings Sync on its own, but for it to properly work you need an extension that provides auth (for Gitpod on the web this is preinstaled, the same way GitHub auth is preinstalled for Desktop VS Code) -> you have to install our extension manually for settings sync for Desktop, regardless if you are developing locally or using a Gitpod backend.

@filiptronicek It worries me a bit that such long explanation is required after reading docs.

@andreafalzetti Do you see how we can rephrase something in document to make points clear?

akosyakov (approved):
lgtm, but I know a lot about Settings Sync

I would like to make sure that it is clear for @andreafalzetti @loujaybee @felladrin ?

felladrin (approved):
Very well written! All clear to me.

PR #1943, opened 2022-04-26
JetBrains plugins docs

### LINKED ISSUE #1637: Document JetBrain's IDE plugin (e.g. client vs server)
Include: 
* What the plugin can do (e.g. it doesn't work from within the IDE right now)
* Include how to update the plugin (screenshot) if issues
* Where to report issues, etc.

akosyakov:
## Description

This PR covers different kind of plugins for each JB product, when to use each and how to install it with detail guidance.

## Related Issue(s)

fix #1637

## How to test

- [Install plugins for IntelliJ IDEA](https://deploy-preview-1943--gitpod-kumquat.netlify.app/docs/ides-and-editors/intellij#install-plugins)
- [Install plugins for GoLand](https://deploy-preview-1943--gitpod-kumquat.netlify.app/docs/ides-and-editors/goland#install-plugins)
- [Install plugins for PhpStorm](https://deploy-preview-1943--gitpod-kumquat.netlify.app/docs/ides-and-editors/phpstorm#install-plugins)
- [Install plugins for PyCharm](https://deploy-preview-1943--gitpod-kumquat.netlify.app/docs/ides-and-editors/pycharm#install-plugins)

## Release Notes

```release-note
NONE
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1943"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1943"></a>


gitpod/docs/ides-and-editors/goland.md
```diff
@@ -41,3 +41,68 @@ Gitpod enables code editing using JetBrains {title} currently via [JetBrains Gat
...
+
+#### Install for repository
+
+You can share a remote plugin with everybody working on the repository by adding it to [.gitpod.yml](/docs/config-gitpod-file) and pushing to your Git repository.
+Each workspace is preconfigured with plugins from this configuration file, here is an example:
+
+```yaml
+jetbrains:
+  goland:
+    - pluginId
```
felladrin, line 97: I guess it would make it clearer for the reader if we list more then one plugin in this dummy list, just to indicate that it's also possible to install two or more plugins to an IDE through this config.

Otherwise, I think we need to explicitly mention it somewhere in this text block.

felladrin (commented):
Great addition to the docs!

I left just a couple of suggestions.

gitpod/docs/ides-and-editors/goland.md
```diff
@@ -41,3 +41,68 @@ Gitpod enables code editing using JetBrains {title} currently via [JetBrains Gat
...
 **Important:** You must restart any started workspaces for your IDE preferences to take effect.
+
+## Install Plugins
+
+Plugins extend the core functionality of {title}:
+
+- Integration with issue trackers, build management servers, and other tools.
+- Coding assistance support for various languages and frameworks.
+- Shortcut hints, live previews, File Watchers, and so on.
+- Coding exercises that can help you to learn a new programming language.
```
loujaybee, line 52: Thought: 

I like the idea of having this context, but I'm wondering if we should link to JetBrains docs regarding benefits/purpose of plugins and keep our reference docs purely about plugin setup 🤔

gitpod/docs/ides-and-editors/goland.md
```diff
@@ -41,3 +41,68 @@ Gitpod enables code editing using JetBrains {title} currently via [JetBrains Gat
...
+
+{title} plugins can either be installed on your **local machine** or in a **remote Gitpod workspace**.
```
loujaybee, line 54: It'd be good if we're able to explain why the difference / what the difference is between the remote and local plugins and their limitations. I've not seen that JetBrains had this documented well already somewhere, but I can take another look.

gitpod/docs/ides-and-editors/goland.md
```diff
@@ -41,3 +41,68 @@ Gitpod enables code editing using JetBrains {title} currently via [JetBrains Gat
...
+</figure>
+For more information on regular plugin management, refer to <a href="https://www.jetbrains.com/help/go/managing-plugins.html" target="_blank">{title} docs</a>.
+
+### Install in Gitpod workspace
+
+Remote plugins require access to remote filesystem and tools, e.g. contributing support of languages and frameworks. Currently a remote plugin can either be installed for the current Gitpod workspace or for repository.
+
+#### Install for the current workspace
+
+You can install a remote plugin only for you for the current workspace following these steps:
```
loujaybee, line 78:
```suggestion
You can install a remote plugin only for your current workspace following these steps:
```

gitpod/docs/ides-and-editors/goland.md
```diff
@@ -41,3 +41,68 @@ Gitpod enables code editing using JetBrains {title} currently via [JetBrains Gat
...
+jetbrains:
+  goland:
+    - pluginId
+```
+
+You can find **pluginId** on the [JetBrains Marketplace](https://plugins.jetbrains.com) page:
+
+- find a page of the required plugid;
+- select the **Versions** tab;
+- and click any version to copy **pluginId** (short name such as `org.rust.lang`) of the plugin you want to install.
```
loujaybee, line 104:
```suggestion
- Find a page of the required plugin
- Select the **Versions** tab
- Click any version to copy the **pluginId** (short name such as `org.rust.lang`) of the plugin you want to install.
```

gitpod/docs/ides-and-editors/goland.md
akosyakov: yes, let's simplify, link and just discuss a difference in our context 👍

gitpod/docs/ides-and-editors/goland.md
akosyakov: I think we can duplicate it in both places. Depends there user will navigate, but it does not harm to repeat ourself.

gitpod/docs/ides-and-editors/goland.md
akosyakov: ok, will put couple plugins, do you have an idea which can be applied to all products and good?

loujaybee:
Looks good to me, Anton. 

I left a few comments on typos etc, two (bigger things) I was thinking about when reading, that I don't have a specific suggestion for, but I'd also be happy to raise a subsequent PR to address, or raise a PR into this one for review, etc. 

1. How we can make it more clear that updating plugins manually doesn't store them (and should be added to configurations), e.g. there's a choice between adding to the YML vs adding manually. 
2. How we can make it more clear about the differences between remote and local extensions. Or more specifically why they're different. I think a small section before we get into the docs could help here.

gitpod/docs/ides-and-editors/goland.md
loujaybee: Yeah. In which case I'd suggest just to pull up some short description of the difference from the introductions of each section up here, so the user can see side-by-side the difference. 

Suggestion 🙏 

```suggestion
{title} plugins can either be installed in two ways:

 1. **local machine** - These changes apply to all workspaces that the user connects to. 
 2. **Remote Gitpod workspace** - These changes apply to all users who connect to the workspace only. 
 
 More details on the different approaches are given below. 

```

gitpod/docs/ides-and-editors/goland.md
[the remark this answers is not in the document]
akosyakov: I meant a user current workspace here, not any other.

akosyakov:
> How we can make it more clear that updating plugins manually doesn't store them (and should be added to configurations), e.g. there's a choice between adding to the YML vs adding manually.

@loujaybee I'm not sure that we should be too pushy here. We configured JB products in such way that plugins do survey restarts. So if a user installs a plugin in one workspace and just use this workspace it will work. If they start another workspace and miss the plugin I would assume that they will consult doc again. I'm not sure that pushing them ahead of time to second approach is sensible.

gitpod/docs/ides-and-editors/goland.md
felladrin: Sure, here are some for general use:

- [Prettier](https://plugins.jetbrains.com/plugin/10456-prettier)
- [BashSupport](https://plugins.jetbrains.com/plugin/4230-bashsupport)
- [IdeaVim](https://plugins.jetbrains.com/plugin/164-ideavim)
- [.​ignore](https://plugins.jetbrains.com/plugin/7495--ignore)

More famous plugins with multi-IDE support can be found in the [Top Downloaded Plugins](https://plugins.jetbrains.com/search?orderBy=downloads&products=androidstudio&products=appcode&products=clion&products=dataspell&products=dbe&products=go&products=idea&products=idea_ce&products=idea_edu&products=mps&products=phpstorm&products=pycharm&products=pycharm_ce&products=pycharm_edu&products=rider&products=ruby&products=webstorm) section.

akosyakov:
@loujaybee @felladrin Could you have a look again? I restrained from usage of local vs remote plugins, since JB docs don't use it. And instead aligned more with them, i.e. installed on JetBrains client or IDE backend. 

I added an introduction section which explains all possible installation targets, their relationships and ways to install/configure plugins.

generators/ide/jb-product.md
```diff
@@ -46,3 +46,65 @@ Gitpod enables code editing using JetBrains {title} currently via [JetBrains Gat
...
+</figure>
+
+#### Preconfigure for repository
+
+You can share a plugin on **{title} backend** with everybody working on the repository by adding it to [.gitpod.yml](/docs/config-gitpod-file) and pushing to your Git repository.
+Each workspace is preconfigured with plugins from this configuration file, here is an example:
+
+```yaml
+jetbrains:
+  { productId }:
```
akosyakov, line 97: rrrr 😬

felladrin (approved):
It looks even better now!

PR #1948, opened 2022-04-26
Fix some details about VS Code Extensions

filiptronicek:
## Description
There are some problems on the page, including: (copied from some feedback I wrote a couple months ago):
- Inconsistencies about OpenVSX, in the docs it is stated that you cannot download a vsix file and manually install it in Gitpod, but AFAIK that’s not the case, it’s not direct usage, it’s just getting the user-uploaded package.
- The docs also mention that Gitpod uses Open VSX, which it does, but it’s only half of the truth. Only one if its integrated IDEs - VS Code Web does, which could make some users confused because if they use our desktop VS Code integration, they use Microsoft Marketplace.
- We should also note there that if the extension they want to get published to OpenVSX [[1](https://github.com/open-vsx/publish-extensions/blob/ccb763e47dd82ed76b5c2df4a5c85c509c406949/lib/reportStat.js#L33-L62)] is from Microsoft / GitHub, they are most likely out of luck and should not be spamming MS with requests to publish Liveshare or Copilot there. We should highlight these extensions which we cannot publish somewhere, I made a list 👀 [[1](https://github.com/open-vsx/publish-extensions/blob/ccb763e47dd82ed76b5c2df4a5c85c509c406949/lib/reportStat.js#L33-L62)]

## Related Issue(s)

This PR is blocked by https://github.com/gitpod-io/website/pull/1914, please merge it before this PR or else some links will break.

## How to test
Read the docz

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1948"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1948"></a>


gitpod/docs/ides-and-editors/vscode-extensions.md
```diff
@@ -62,10 +60,10 @@ You can view all pre-installed extensions by navigating to VS Code's Extensions
...
+If you cannot find an extension by searching in Gitpod using [VS Code Web](vscode-browser), it probably means that the extension hasn't been added to the [Open VSX](https://open-vsx.org/) registry yet.
 
 In that case, please reach out to the extension author and politely ask them to publish their extension to the vendor-neutral, open source Open VSX registry. The "[How to Publish an Extension](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions)" docs provide step-by-step instructions.
 
-Please note that `.vsix` files downloaded from the [Visual Studio Marketplace](https://marketplace.visualstudio.com/vscode) should not be installed in Gitpod, because Microsoft prohibits the direct use of their marketplace by any non-Microsoft software, even though most extensions are actually open source and not developed or maintained by Microsoft.
+Due to legal reasons, there are some extensions we just can't publish. These include some popular proprietary extensions like GitHub Copilot or the Remote extensions from Microsoft.
```
akosyakov, line 67: There is not legal reasons prohibiting MS publishing to OpenVSX themself. I think we should rather encourage reaching out to show demand for that to extension authors. I won't call requests from different users spamming.

akosyakov (approved):
looks good besides https://github.com/gitpod-io/website/pull/1948/files#r859439026

akosyakov:
Settings Sync doc should be deployed and doc merged before landing this one, correct?

filiptronicek:
> Settings Sync doc should be deployed and doc merged before landing this one, correct?
Exactly, @akosyakov, I'll update the PR description to indicate that this is blocked by it.

gitpod/docs/ides-and-editors/vscode-extensions.md
akosyakov: @filiptronicek Could you revert this change please? 🙏

filiptronicek: @akosyakov do we keep the fact that users should not use `.vsix`es from MS?

akosyakov: yes, it is still true, no?

filiptronicek: I think not. I just read the ToS again and if a user obtains it via the Marketplace online interface, it doesn't specify what the user can do with the files.

akosyakov: It is not really important what we think about ToS, we are not lawyers and don't really konw about MS intent here could you please revert it and put to previous state?

filiptronicek: Done in https://github.com/gitpod-io/website/pull/1948/commits/7495caddbc0892974a9a18155f5bb3300f9560a8

akosyakov:
@filiptronicek let's merge 🚀

PR #1973, opened 2022-04-28
Fix GKE certificate statement

csweichel:
## Description
Minor correction regarding custom CA certs on GKE - we are not currently investigating a solution with Google. Google has confirmed that such a solution is not on the horizon.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1973"></a>


corneliusludmann:
cc @lucasvaltl

corneliusludmann (approved):

PR #1980, opened 2022-05-02
[self-hosted] Link to cluster requirements

csweichel:
## Description
Links to the cluster requirements when talking about clusters, rather than to the components that run on it.
This may be a stop-gap fix and we might want to harmonise the cluster requirements and required services page. I.e. split requirements from the setup guide.

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/1980"></a>


corneliusludmann (approved):
Definitely. Thanks for fixing it!

PR #2000, opened 2022-05-09
Create Changelog

Cahllagerfeld:


atduarte:
A lot of these entries seem irrelevant to our users. I'm unsure what the guideline is regarding those but left a few comments.

src/lib/contents/changelog/2022-05-09.md
```diff
@@ -0,0 +1,55 @@
...
+date: 2022-05-09
+image: 2022-05-09.jpg
+alt: TODO
+---
+
+<script>
+  import Contributors from "$lib/components/changelog/contributors.svelte";
+</script>
+
+TODO: Insert the featured changelog entry here.
```
loujaybee, line 13:
```suggestion
Gitpod is built to be flexible. It’s important that our users can develop with the tools that make sense for their current task and use the tools they are most comfortable with. We “integrate, don’t dictate”. This design principle extends also to our [[IDE and editor integrations](https://www.gitpod.io/docs/ides-and-editors)](https://www.gitpod.io/docs/ides-and-editors) where we have multiple options in the browser and on desktop both VS Code and JetBrains (also see: [[our partnership with JetBrains](https://www.gitpod.io/blog/gitpod-jetbrains)](https://www.gitpod.io/blog/gitpod-jetbrains)). 

We encourage Gitpod users to use [[one workspace per task](https://www.gitpod.io/docs/workspaces#ephemeral-vs-long-lived)](https://www.gitpod.io/docs/workspaces#ephemeral-vs-long-lived), move their development configurations to code, and embrace ephemerality to really get the most out of Gitpod. However, those users who choose to swap between VS Code Desktop and VS Code in the browser for editing code will have noticed that their VS Code settings (themes, plugins, and fonts, etc) weren’t kept in sync. 

But since we know just how important it is for our users to have their setups and configurations just as they like them, we’re excited to let you know that [[**VS Code Settings Sync](https://www.gitpod.io/docs/ides-and-editors/settings-sync)](https://www.gitpod.io/docs/ides-and-editors/settings-sync) now works between VS Code Desktop and the browser in Gitpod** 

**Note:** VS Code browser settings are already synced by default. 

To enable VS Code Settings Sync on your desktop: 

1. Open your Gitpod workspace in VS Code Desktop
2. Use the command palette to select “Settings Sync: Enable signing in with Gitpod”
3. Restart your VS Code Desktop
4. Resolve any settings conflicts between your browser and desktop

And you’re set! All your theme, plugins, and other settings will be synced in real-time and be reflected also when you’re working in VS Code in the browser. 

Check out the [[VS Code settings sync documentation](https://www.gitpod.io/docs/ides-and-editors/settings-sync)](https://www.gitpod.io/docs/ides-and-editors/settings-sync) for more details.
```

PR #2005, opened 2022-05-09
Fix Changelog

Cahllagerfeld:


Cahllagerfeld (commented):
Hey @nisarhassan12, I fixed the blurry Changelog-Header images and made them consistent in size. Although being a bit bigger we're always below 100kb, which is fair I think. 
I used [MozJpeg](https://mozjpeg.com/) for optimizing the images

chrifro:
Nice, looks great @Cahllagerfeld 👏

Small nit, the SSH image is a bit complex and it's difficult to understand what's the most important information here.
Alternatively, I created this image. WDYT?
Paste

PR #2010, opened 2022-05-09
Optimize some homepage images for modern browsers

pavelloz:
## Description
* Optimized some images in place, to reduce their size - should make it faster for most (70% general population, probably 90%+ for devs) people
* Generated avif versions for some png images (IDEs) for modern browsers
* Used picture tag so browser knows when to use fallback image
* Reworked the link preloads to preload whats visible from the get go, and eager load (using native loading=eager) the rest of screenshots - should save some bandwidth on mobile
* Lazyload avatars in testimonials (and decode asynchronously) - far below the fold + some of them are hidden behind horizontal scroll - should make a positive difference on mobile

## How to test
* Go to homepage
* See if IDEs screenshots still work
* See if avatars in testimonials are present

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2010"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2010"></a>


nisarhassan12:
Hi, @pavelloz Thanks for the PR 🧡 . 

It's great that you have added the `.avif` version of the images that will cover the majority of our users 👍🏿 and made the `.pngs` as the fallbacks.

Your changes do significantly reduce the initial payload for the browser and unnecessary images don't get preloaded(i.e no unnecessary data is being shipped around) unless required. 

I tried out the PR preview and the following is what the loading experience looks like for PR: 

https://user-images.githubusercontent.com/46004116/167448772-2e1af2b5-b7f8-479f-897e-e501671774d5.mov

It looks like your changes did improve the performance but removing preloading and adding eager loading to the avatars sort of have a visual impact and it feels as if the perceived performance has decreased a bit i.e the screenshots take a few seconds to load when an IDE logo is being hovered upon whereas with preloading it is almost instant. 

For the avatars, the images are rendered as the content is being brought into the viewport which is not that pleasant to the eyes cc @chrifro @Cahllagerfeld What do you guys think?

filiptronicek:
@nisarhassan12 @pavelloz maybe we could automate all of this with a build step and a custom image element? 

The image component would just have multiple file types and we could only specify one if we wanted to.

We could use the [Squoosh CLI](https://github.com/GoogleChromeLabs/squoosh/tree/dev/cli) for optimizing images as part of our build step as well, meaning we don't have to have multiple versions of images in our repo.

pavelloz:
@nisarhassan12
Hmmm, I guess localhost:3000 was too fast to test it properly :)

One way (and the only I can think of) of dealing with that regression would be to switch back to preload in link tags, but then preloading both formats is not ideal. Custom element could fix that.

As for avatars, maybe just resizing them to 48x48 would be enough, since they are displayed at 48x48 on the page.

BTW. Is firefox popular within your viewers

@filiptronicek
It is one way of doing it, but the build time would probably explode (optimizing images can be very time consuming), and without proper caching of generated images and detection from Sqoosh if the work is needed, it would be repeated on every build.

Custom element would be great for the preload duplication problem though - it could detect which image to preload in the head and load as an img source. 

---

Having CDN (I can only recommend cloudinary) that serves the best image format would be ideal.
I just wanted to throw something together quick to see the reaction for this type of PR - sometimes there is zero interest in performance :)

---

Ive sent 2nd commit, this should move things in a better direction.

filiptronicek:
>  It is one way of doing it, but the build time would probably explode (optimizing images can be very time consuming), and without proper caching of generated images and detection from Sqoosh if the work is needed, it would be repeated on every build.

Agreed, I guess we could then generate them and store them in the repo, maybe have a GitHub Action for generating missing ones and if there are changes commiting them back. 

Another alternative may be worth considering (could be a very crazy idea :) is to use a proxy like https://images.weserv.nl/. I have been using them for quite some time, and it should be easy to [self-host](https://github.com/weserv/images/tree/5.x/docker) as well. It's super fast and serves our desired formats without any special files on the fly.

pavelloz:
^ Yes. Having it done on edge (or using dedicated CDN like cloudinary) would be great, because of its transparency, lack of manual labor and cleaningness of repository. I even wrote that in my comment, but deleted to not distract anybody :)

nisarhassan12:
> Ive sent 2nd commit, this should move things in a better direction.

Thanks @pavelloz Your 2nd commit did make things a lot better. 

But IDE screenshots are no longer appearing in on Safari(15.3) see the below video:

https://user-images.githubusercontent.com/46004116/167561770-cf90a258-4236-454c-b79f-2283f5f2095c.mov

Where possible we need to make sure that the experience is equally good on Safari and IOS devices because many of our users probably use IOS devices and both Chrome and Firefox there use Webkit under the hood and `.avif` is not supported there yet. I'm not sure why didn't the PNGs as fallback worked. 

Even if the PNGS worked they won't be preloaded making the experience a bit degraded for IOS users as compared to others. 

> Custom element would be great for the preload duplication problem though - it could detect which image to preload in the head and load as an img source.

 Let's maybe use a custom element as you suggested above 🙂 or else just use optimized PNGs with preload. 
 
 > As for avatars, maybe just resizing them to 48x48 would be enough, since they are displayed at 48x48 (or 48 * 2 for retinas) on the page.
 
 Let's go with 96x96 for retinas :) 
 
 > @nisarhassan12 @pavelloz maybe we could automate all of this with a build step and a custom image element?
 
 On the old [website repo](https://github.com/gitpod-io/retired-gatsby-website), we had https://github.com/apps/imgbot which used to raise PRs whenever some new images get added to the repo. I'll reach out to our team internally to re-enable that.

filiptronicek:
> On the old [website repo](https://github.com/gitpod-io/retired-gatsby-website), we had [github.com/apps/imgbot](https://github.com/apps/imgbot) which used to raise PRs whenever some new images get added to the repo. I'll reach out to our team internally to re-enable that.

@nisarhassan12 ImgBot would be nice to add, but I was thinking more of adding the modern image formats **alongside** the pngs and jpegs. Adding WebPs and AVIF versions of images is I think not possible with the bot.

pavelloz:
I extracted whatever is working to another PR, this wont go anywhere it seems.

Issue #1893, opened 2022-04-11
Toggle


SaraHesse:
[[Figma Link](https://www.figma.com/file/7cegM2Pg34OpJcOmrjbijj/Website-2022?node-id=286%3A7601)](https://www.figma.com/file/7cegM2Pg34OpJcOmrjbijj/Website-2022?node-id=286%3A7601)

-  [x] Adjust height accordingly to the Figma design file (48px instead of 60px height)
- [x] New hover state (outline)
- [x] Dark mode

chrifro:
The toggle component has been updated in the component library. Could you adjust it on the website as well @nisarhassan12 🙏

Changes: the hover has a linear stroke now, there is no more shadow, and the selected item changes on hover. The button height was never in sync with the height in the component.


Thanks!

nisarhassan12:
> - [ ] and the selected item changes on hover.
@chrifro I have made the changes in https://github.com/gitpod-io/website/pull/2050 apart from changing the selected item on hover. I'll make a follow-up Pr for that.

PR #2050, opened 2022-05-16
toggle adjustments.

nisarhassan12:
Addresses parts of #1893 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2050"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2050"></a>


nisarhassan12:
@chrifro [here](https://www.figma.com/file/7cegM2Pg34OpJcOmrjbijj/Website-2022?node-id=286%3A7601) in the mockup provided by MAD for the dark mode they are using an orange which is not in the colour palette. Should I use the new colour that they suggested for the dark mode or the existing orange for the dark is fine? Thanks 

image

chrifro:
>Should I use the new colour that they suggested for the dark mode or the existing orange for the dark is fine?

You were looking at an early draft from MAD. Please use the Kumquate Ripe color as shown [here](https://www.figma.com/file/7cegM2Pg34OpJcOmrjbijj/Website-2022?node-id=797%3A7038) 

>Also, the live site uses a white background for the switch which has better contrast than using off-white as suggested in the new mockup. What if we make the gradient stroke appear without having to hover?

Great suggestion. Let's do that

>Also, the switch is just a checkbox styled like a switch it is not common on the web to check/uncheck the checkboxes on hover i.e checkboxes generally work with clicks. What is the reason we are opting to make the switch change on hover :)?

It was MAD's idea. I agree with you. Let's keep the hover style how we currently have it :)

PR #2052, opened 2022-05-16
Adjust pricing cards

### LINKED ISSUE #1880: Hovering card: Remove Glow
**Problem:** 

While hovering a card, the glow gives the impression that a user can interact with the card (click it), which is not the case. It’s the same effect which is used to highlight one of the cards, so it’s confusing.

Bildschirmfoto 2022-04-05 um 13 45 22

**Task:**

We’d recommend to remove the glow from hovering over the card.

nisarhassan12:
Fixes #1880


chrifro (commented):
The card stroke is missing otherwise it LGTM

PR #2075, opened 2022-05-18
Adapt theme to system theme changes and make system default theme option more accessible

### LINKED ISSUE #2064: Adapt theme to system theme changes and make system default theme option more accessible
### Problem to solve

Following up from new awesome dark theme support added for `www.gitpod.io`:
1. The theme does not adapt when system theme changes, which is what happens when system theme is set to automatically change on sunset or specific times.
2. The interaction of the theme preference in the footer seems broken when the system theme matches the current theme.

### Proposal

We could try the following:
1. Adapt theme to system theme changes.
2. Make the system default theme option more accessible by having an active state on the footer theme switch.

nisarhassan12:
Fixes #2064 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2075"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2075"></a>


gtsiolis (commented):
LGTM, @nisarhassan12! 🔮 

However, this will now default to light theme instead of the system default. Do we want this change? 🌔

nisarhassan12:
> - [x]  However, this will now default to light theme instead of the system default. Do we want this change? 🌔

Thanks for the review @gtsiolis I have changed that to have "system" selected as the default option.

PR #2079, opened 2022-05-19
make languages and frameworks discoverable

akosyakov:
## Description

Right now language are hidden from a user and one should use quick search properly to find anything useful. 
This PR move them to top level.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


Cahllagerfeld:
I don't know the historic context of these (maybe @ghuntley can help here) but all of these pages contain kind of a "deprecation disclaimer" and point to the quickstarts as source of truth 😅
grafik

akosyakov:
@Cahllagerfeld I don't consider this page as a quick start but instead would like to turn each language/framework page as an entry point for developers, i..e have a sub section for each import point of their journey in Gitpod, i.e. how to configure dev tools (maven/gradle) with Dockerfile in context of Gitpod, configure IDEs, run/debug/profile, work with databases, application servers and frameworks.

I also don't see on Quicstart a guide about Java at all and alll guides have same content without particularies of languages/frameworks. Is it fine to remove dislaimer?

Cahllagerfeld:
I have the same opinion as you, I think the pages are containing more relevant information and tips (e.g on how to debug) than the current quickstarts. Because of that, I'd be personally fine with removing the disclaimer. 

The only thing I'm not quite sure about is the reason why we decided to add this disclaimer and because of that if there is something we aren't considering currently. (I believe there was an internal discussion earlier with this outcome before adding the disclaimer)

akosyakov:
@Cahllagerfeld I deleted disclaimers.

src/lib/contents/docs/menu.ts
```diff
@@ -64,6 +64,28 @@ export const MENU: MenuEntry[] = [
...
+    M("Rust", "languages/rust"),
+    M(".NET", "languages/dotnet"),
+    M("Dart", "languages/dart"),
+    M("Julia", "languages/julia"),
+    M("LaTeX", "languages/latex"),
+    M("R", "languages/r"),
+    M("Kotlin", "languages/kotlin"),
+    M("Pandas", "languages/python#pandas"),
+    M("Deno", "languages/deno"),
+  ]),
```
Cahllagerfeld, line 88: would it make sense to sort them alphabetically, by popularity, or are they already sorted by a schema i don't know? 😅

loujaybee, line 88: A couple of slightly alternate options for consideration. 

|A - Current| B - configure by language| C - languages|
|-|-|-|
|-|-|-|
|image | image|image|

**A** - Shown in current position / labelling. 
**B** - Update to mention "configuring", so it's more aligned with the configure section. Is more explicit about the purpose of the section.
**C** - Drop frameworks entirely (as we currently don't have one listed), with the potential to add a new section, or extend in the future. The languages section is already quite long—do we realistically see ourselves adding X more frameworks to the list?

Cahllagerfeld: My 2 cents: 
I'd go with Option C, technically the .gitpod.yml-File for node projects looks nearly always the same, so here it's technically no difference between Framework & Language. A downside for this would be that the JS-World thinks in Frameworks and not mentioning `React` explicitly might lead to the assumption that React isn't supported

akosyakov: I think they are already sorted somewhat by popularity.

akosyakov: let's go with Languages

I think we will need to group and rework guides anyway. Frameworks can go under languages.

akosyakov: dropped frameworks

akosyakov:
@loujaybee @Cahllagerfeld Could you have a look again please?

loujaybee (approved):

PR #2083, opened 2022-05-19
Fix relicts from old Toc

Cahllagerfeld:


loujaybee (approved):
Seems to fix the issue 🙏

PR #2085, opened 2022-05-19
Update SSH documentation

jeanp413:
## Description

Updates SSH documentation

Related https://github.com/gitpod-io/openvscode-server/pull/346

## Release Notes

```release-note
NONE
```


gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,21 @@ To get started:
...
-VS Code desktop support installs, manages and runs the [Local Companion](/docs/ides-and-editors/local-companion) app under-the-hood so you don't have to. However, if you're currently using the local companion, you can can continue to use it independently.
-
 <sup>1</sup> Yes this also means you get to keep your keyboard shortcuts just the way you prefer them. You know, Ctrl/Cmd + W to close an editor tab 😉.
 
+## Remote development with SSH
+
+There are two modes to connect remotely to a workspace, both of them use the `Remote - SSH` extension:
+
+- **SSH Gateway**, allows you to connect directly to your workspace (it needs to be configured in your Gitpod self-hosted installation).
+- **Local App** (`deprecated`), uses a separate program binary that tunnels SSH traffic .
```
akosyakov, line 33: `Local Companion` 

local app is internal term

akosyakov: We should make it clear that using the local companion deprecated, but not the local companion itself.

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,21 @@ To get started:
...
+
+As Local App has a lot of drawbacks and is unstable, it is now deprecated in favor of SSH Gateway.
```
akosyakov, line 35: The local companion as standalone is fine, but using it for integrations in VS Code turned out to be problematic. We should frame it properly here.

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,21 @@ To get started:
...
+
+By default the [Gitpod extension](https://marketplace.visualstudio.com/items?itemName=gitpod.gitpod-desktop) will try to connect using SSH Gateway but if it's not configured yet, it will fallback to using Local App.
+
+If you have SSH Gateway set up but can't connect due to other factors like a firewall blocking SSH connections, you can force to use Local App in the meantime by configuring the `gitpod.remote.useLocalApp` setting.
```
akosyakov, line 39: We need to point out that this settings will be removed in future and issues with SSH Gateway should be reported to us using troubleshooting docs for VS Code Desktop.

jeanp413: You mean just a link to `/docs/troubleshooting` or another new section in there? (I did the former)

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,23 @@ To get started:
...
 <sup>1</sup> Yes this also means you get to keep your keyboard shortcuts just the way you prefer them. You know, Ctrl/Cmd + W to close an editor tab 😉.
 
+## Remote development with SSH
+
+There are two modes to connect remotely to a workspace, both of them use the **Remote - SSH** extension:
+
+- **SSH Gateway**, allows you to connect directly to your workspace (it needs to be configured in your Gitpod self-hosted installation).
+- **Local Companion**, uses a separate program binary that tunnels SSH traffic.
+
+As using _Local Companion_ to tunnel SSH traffic has a lot of drawbacks and is not reliable, it is now `deprecated` in favor of SSH Gateway (_Local Companion_ itself is not deprecated and can be used standalone).
```
akosyakov, line 35: `has drawbacks` is enough

akosyakov (approved):
nice, let's coordinate next week on when what to merge and release

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,23 @@ To get started:
...
+
+By default the [Gitpod extension](https://marketplace.visualstudio.com/items?itemName=gitpod.gitpod-desktop) will try to connect using SSH Gateway but if it's not configured yet, it will fallback to using Local Companion.
+
+If you have SSH Gateway set up but can't connect due to other factors like a firewall blocking SSH connections, you can force to use Local Companion in the meantime by configuring the `gitpod.remote.useLocalApp` setting. Keep in mind this setting will be removed in the future along with Local Companion mode.
+
+If you find any issues using SSH Gateway, please [report them](/docs/troubleshooting).
```
akosyakov, line 41: `docs/troubleshooting#gitpod-logs-in-vs-code-web`

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,23 @@ To get started:
...
 
-VS Code desktop support installs, manages and runs the [Local Companion](/docs/ides-and-editors/local-companion) app under-the-hood so you don't have to. However, if you're currently using the local companion, you can can continue to use it independently.
-
 <sup>1</sup> Yes this also means you get to keep your keyboard shortcuts just the way you prefer them. You know, Ctrl/Cmd + W to close an editor tab 😉.
 
+## Remote development with SSH
+
+There are two modes to connect remotely to a workspace, both of them use the **Remote - SSH** extension:
+
+- **SSH Gateway**, allows you to connect directly to your workspace (it needs to be configured in your Gitpod self-hosted installation).
```
akosyakov, line 32: @iQQBot @corneliusludmann  Can we link here to some self hosted docs about SSH Gateway?

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,23 @@ To get started:
...
+- **Local Companion**, uses a separate program binary that tunnels SSH traffic.
```
akosyakov, line 33: let's link **Local Companion** to `/docs/ides-and-editors/local-companion`

gitpod/docs/ides-and-editors/vscode.md
akosyakov: `_Local Companion_ itself is not deprecated and can be used standalone` can be just an additional sentence without parenthesis

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,23 @@ To get started:
...
+- **SSH Gateway**, allows you to connect directly to your workspace (it needs to be configured in your Gitpod self-hosted installation).
+- [**Local Companion**](/docs/ides-and-editors/local-companion), uses a separate program binary that tunnels SSH traffic.
+
+As using _Local Companion_ to tunnel SSH traffic has drawbacks and is not reliable, it is now `deprecated` in favor of SSH Gateway. Local Companion itself is not deprecated and can be used standalone.
+
+By default the [Gitpod extension](https://marketplace.visualstudio.com/items?itemName=gitpod.gitpod-desktop) will try to connect using SSH Gateway but if it's not configured yet, it will fallback to using Local Companion.
+
+If you have SSH Gateway set up but can't connect due to other factors like a firewall blocking SSH connections, you can force to use Local Companion in the meantime by configuring the `gitpod.remote.useLocalApp` setting. Keep in mind this setting will be removed in the future along with Local Companion mode.
+
+If you find any issues using SSH Gateway, please [report them](/docs/troubleshooting#gitpod-logs-in-vs-code-web-and-desktop).
```
jeanp413, line 41: test

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,23 @@ To get started:
...
 <sup>1</sup> Yes this also means you get to keep your keyboard shortcuts just the way you prefer them. You know, Ctrl/Cmd + W to close an editor tab 😉.
 
+## Remote development with SSH
+
+There are two modes to connect remotely to a workspace, both of them use the **Remote - SSH** extension:
+
+- **SSH Gateway**, allows you to connect directly to your workspace (it needs to be configured in your Gitpod self-hosted installation).
+- [**Local Companion**](/docs/ides-and-editors/local-companion), uses a separate program binary that tunnels SSH traffic.
+
+As using _Local Companion_ to tunnel SSH traffic has drawbacks and is not reliable, it is now `deprecated` in favor of SSH Gateway. Local Companion itself is not deprecated and can be used standalone.
```
akosyakov, line 35: please replace with `As using _Local Companion_ to tunnel SSH traffic has drawbacks,`

gitpod/docs/ides-and-editors/vscode.md
```diff
@@ -23,10 +23,23 @@ To get started:
 
 This command will open your local VS Code, connect to the Gitpod workspace and let you develop, test & debug your application as if the code was running locally.
 
-VS Code desktop support installs, manages and runs the [Local Companion](/docs/ides-and-editors/local-companion) app under-the-hood so you don't have to. However, if you're currently using the local companion, you can can continue to use it independently.
-
 <sup>1</sup> Yes this also means you get to keep your keyboard shortcuts just the way you prefer them. You know, Ctrl/Cmd + W to close an editor tab 😉.
 
+## Remote development with SSH
```
loujaybee, line 28: **Praise:** Thanks a lot for picking this up, JP! 

**Suggestion:** I think presenting this as two options risks being misinterpreted as both options being valid (rather than SSH gateway as the primary support method, with local companion as a fallback). I'd suggest to keep only the SSH Gateway content on this page, and separate out the local companion fallback, e.g. 

1. **Keep the SSH Gateway documentation in this page, e.g. "How VS Code Desktop connects via SSH"** - Document the setup/considerations for SSH Gateway as if the local companion approach option doesn't exist. 
2. **Move docs on the fallback of local companion to an [FAQ](https://deploy-preview-2085--gitpod-kumquat.netlify.app/docs/ides-and-editors/faqs), e.g. "VS Code Desktop won't connect via SSH"** - Where we detail the local companion fallback, and make the note of deprecation, referring to the main method in the main docs. 

@jeanp413 - Happy to discuss further or even push a commit with some changes if that would help 🙏

gitpod/docs/ides-and-editors/vscode.md
jeanp413: Do you mean adding the comma after `drawbacks`?

gitpod/docs/ides-and-editors/vscode.md
jeanp413: I think this would make sense if connecting with Local App is completely optional, i.e. user always has to manually update `gitpod.remote.useLocalApp` setting but right now if ssh gateway is not configured it will automatically fallback to Local App so it's still a valid option.
Maybe we should set a date when we remove the automatic fallback and update the docs and another date to completely remove local app mode

jeanp413: cc @akosyakov

akosyakov: @jeanp413 I agreed, and don't think the first option is viable. We need to keep and document connectivity with local app from VS Code Desktop somewhere, since we are not sure that switching to SSH Gateway won't break some users.

But second point is good. @loujaybee I think the better place will be https://www.gitpod.io/docs/troubleshooting no FAQ.

We cover here as local app does not exist and then add a note at the end if you cannot connect see this troubleshooting docs.

loujaybee (approved):
Merged some changes that I ran through with @akosyakov to fix: 

1. We mention “SSH Gateway” but we provide no link or context about what it is, and it’s not really an industry term/concept like a bastion, etc. I’d love to document it, but we have no reference to point to, I think it raises more questions than it answers…
2. We mention drawbacks of local companion but do not explain what they are. I think we can omit this statement to avoid questions.
3. The language around deprecation of local companion mode is ambiguous. We mention that local companion is not deprecated, but the wording is quite confusing.

Commit: https://github.com/gitpod-io/website/pull/2085/commits/213f3a83f1a306bc1aaae55f5b80afcc9f619947

All good to go for me, @jeanp413 🚀  

Again, thanks a lot for making sure to update the docs 🙏

gitpod/docs/ides-and-editors/vscode.md
loujaybee: Resolved with: https://github.com/gitpod-io/website/pull/2085#pullrequestreview-983287376

🙏

Issue #2024, opened 2022-05-11
Get rid of the Github avatars


nisarhassan12:
Originated from https://gitpod.slack.com/archives/C01KGM9ETU6/p1652282055925879

![image](https://user-images.githubusercontent.com/46004116/167892770-2db7142c-4c10-4679-be94-10eadeec8470.png)

We should get rid of all of the Github avatars on the website from /blog, /changelog, /guides and elsewhere to avoid the addition of cookies from GitHub.

nisarhassan12:
Fixed in https://github.com/gitpod-io/website/pull/2094

PR #2094, opened 2022-05-21
cookies management updates. 

nisarhassan12:


nisarhassan12:
@MircoatGitpod I have removed the save button as the checkboxes add/remove the cookies as they are checked/unchecked. Is this behaviour fine? Or do we need to only save the changes once the save button is clicked? I feel like having the save button is an extra step. WDYT? 
 
image

Also, is it expected to have the Cookie banner being closed when the close settings text is clicked?

MircoatGitpod:
> @MircoatGitpod I have removed the save button as the checkboxes add/remove the cookies as they are checked/unchecked. Is this behaviour fine? Or do we need to only save the changes once the save button is clicked? I feel like having the save button is an extra step. WDYT?
> 
> image
> 
> Also, is it expected to have the Cookie banner being closed when the close settings text is clicked?

@nisarhassan12 I think this is only a matter of user experience, both could be fine. I think the save button gives some kind of a feedback/confirmation to an user about his updated preferences. Without anything some users might not realise how this works are they are used to save buttons. When settings are automatically saved we might want to put some confirmation in place i.e. "Preferences have been updated"

nisarhassan12:
Thanks, @MircoatGitpod I have updated the Pr so that the cookies are updated when the save button is clicked.

nisarhassan12:
@MircoatGitpod The following screenshot shows the current cookies that we have on our Homepage. Only `ajs_anonymous_id` is set from the website itself the rest comes from the Gitpod dashboard. 
Screen Shot 2022-05-30 at 10 08 50 PM

When reject all is pressed we will only reject the cookies that are added from the website? i.e `ajs_anonymous_id`, `gp-analytical` and `gp-targetting`.

MircoatGitpod:
@nisarhassan12 

> When reject all is pressed we will only reject the cookies that are added from the website? i.e ajs_anonymous_id, gp-analytical and gp-targetting.
Technically we should also suppress the code that is loading external cookies then. I believe most of these come from Google/embedded youtube videos. There seems to be a way in embedding youtube videos without any cookies.

src/routes/cookies.md
```diff
@@ -0,0 +1,81 @@
...
+| Website          | yt-remote-session-app            | Targeting/Marketing   | Stores the user's video player preferences using embedded You Tube video                                                                                   | Persistent      | Yes                | https://policies.google.com/privacy?hl=en |
+| Website          | yt-remote-session-name           | Targeting/Marketing   | Stores the user's video player preferences using embedded You Tube video                                                                                   | Persistent      | Yes                | https://policies.google.com/privacy?hl=en |
+| Website          | ajs_user_id                      | Analytics/Performance | Generates a userID to identify reoccurring website visitors                                                                                                | Persistent      | Yes                | https://segment.com/legal/privacy/        |
```
nisarhassan12, line 78: @MircoatGitpod looking at the entire source code for the site there is no cookie being set with the name `ajs_user_id` instead it's a value stored in local storage by Segment JS. Should I remove this from here?

nisarhassan12:
@MircoatGitpod Regarding suppressing all the external cookies we have some cookies coming from GitHub because of the GH avatars on `/blog`, `/changelog`, `/guides` and so on see https://github.com/gitpod-io/website/issues/2024

While totally possible it will take a lot of effort to have all the avatars being hosted on our website i.e it will make the changelog process a bit more lengthy i.e whoever is adding the new changelog will have to add the avatars for any new/external contributors themselves manually.

@Cahllagerfeld maybe we could make a script that runs before the build to fetch all the avatars specified within a file(the file could contain the GitHub usernames) from there we can download the avatars and put them in `/public/avatars` or something similar? WDYT?

Cahllagerfeld:
I think adding a script should work 👍 We're already having `authors.ts` (I believe its called like that), which does exactly that, doesn't it

nisarhassan12:
Thanks, @Cahllagerfeld I think `authors.ts` does that but only for the blog post authors and not for the changelog authors.

MircoatGitpod:
thank you @nisarhassan12! This is looking good from my perspective.

Just fyi - once we are publishing this, we should add the current date to the cookie policy.

@Cahllagerfeld yes, we should get this publish as soon as possible. I will wait with approving this so we can take care about the Github cookies.

Cahllagerfeld:
Nisar, I just spend a bit of time on this: Especially for the changelog, I think having a single file with authors is not very practical, because there we might have the scenario that community members are displayed. We would have to capture ALL contributors and make sure they are on the list and store their images. I think it's overall possible doing this, but I'm not that sure if it's that practical 😅  What do you think about this?

nisarhassan12:
> I think having a single file with authors is not very practical
@Cahllagerfeld yep it is not practical but we have to do this to get rid of the GitHub cookies.

Instead of populating `authors.ts`, I would suggest that we should create a file called `gh-avatars.ts` or something similar with a simple array of GitHub handles like the following:

```ts
const githubSlugs = ['Cahllagerfeld', 'nisarhassan12', 'chifro', 'jankeromnes']
``` 

which can then be used to download the images in `/static/avatars` for every new contributor we just add the GH handle at the end of the array.

src/routes/api/[slug].avatar.ts
```diff
@@ -0,0 +1,10 @@
+import type { RequestHandler } from "@sveltejs/kit";
```
Comment (Cahllagerfeld) [inline src/routes/api/[slug].avatar.ts:1]: Hey @nisarhassan12, I came up with an endpoint that proxies the image download for us. All we need to do is change all occurrences, where GH-Images are getting downloaded to point to that endpoint.

I exchanged it for `avatars` already. That was the only thing I had in mind. If we have any other places, feel free to adjust the request there as well.

src/routes/api/[slug].avatar.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export const get: RequestHandler = async ({ params }) => {
+  const res = await fetch(`https://www.github.com/${params.slug}.png`);
+
+  return {
+    body: res.body,
+    headers: { "Cache-Control": "public, max-age=604800" },
```
Comment (Cahllagerfeld) [inline src/routes/api/[slug].avatar.ts:8]: I added caching to the images, so they don't get reloaded every time someone navigates to the page. It's currently 1 Week. Github-API sets this header to 300 Seconds, which is 5 minutes.

We can adjust this in the future, I think by having more control now, we can adjust this in the future

Cahllagerfeld (commented):
Short explanation what I did
~~Wait its not working with the deployed version~~

src/routes/api/[slug].avatar.ts
Comment (Cahllagerfeld) [reply src/routes/api/[slug].avatar.ts]: This seems to be a  Netlify-Specific thing... again..

src/functions/avatar.ts
```diff
@@ -0,0 +1,19 @@
+import type { Handler } from "@netlify/functions";
```
Cahllagerfeld, line 1: Migrated it now to a Netlify-Endpoint, which is working. Could you please wait with merging. I left the other implementations in the Coding, so we can migrate to an Svevltekit-native Endpoint, once we bump our versions

nisarhassan12: Thanks @Cahllagerfeld When do you think we can get this merged ?

src/routes/api/[slug].avatar.ts
Comment (nisarhassan12) [reply src/routes/api/[slug].avatar.ts:1]: I think avatars are the only place where we have the images coming from GitHub so we should be good to go.

src/functions/avatar.ts
Cahllagerfeld: From my side its done, I was just wondering if we are having any other occurrences, where we load images from somewhere else except `avatar`

MircoatGitpod: thank you @nisarhassan12 @Cahllagerfeld 🙏

@nisarhassan12 Lets adjust the cookie policy to the current date and we can publish it right away. 🚀

MircoatGitpod:
@nisarhassan12 Apart from adding the current date to the cookie policy, we should also put a Cookie Policy Reference/Hyperlink into our Website footer between "Privacy Policy" and "Cookie Preferences"
image

Thats would be everything and we can ship it 🚀

src/routes/cookies.md
nisarhassan12: As suggested I tried to suppress it but the suppression does not work completely and that is because of the fact that it is added to the local storage by the library itself that we don't control. I can look into it in a follow-up Pr. 

@jakobhero Just to make sure that we are all aligned here removing it won't impact the way segment works? right?

benjaoming:
Hello :wave: I'm interested in how consent is managed in general on The Internet. Would anyone reading this with insights into the project mind telling me if this PR stores an auditable trail of consent on your server-side or as a cookie on the client-side? I haven't been able to tell directly from the PR, but I almost definitely missed something because I'm not familiar with this code base. Thanks :pray:

MircoatGitpod:
HI @benjaoming 👋 sure! In our case consent preferences are stored as a cookie on the client-side (one for each category of cookies gp-necessary, gp-analytical, gp-targeting).

benjaoming:
Thanks @MircoatGitpod :pray: I'm not a researcher, but I observe this pattern as clearly dominant, too. I'll try to put something together on digital consent during 2022, some blog post and a reference implementation of the server-side. If you're interested, follow this project: https://github.com/django-denmark/django-consent/

PR #2104, opened 2022-05-23
revert: comment out flaky button

Cahllagerfeld:


Cahllagerfeld (commented):
https://answers.netlify.com/t/missing-headers-in-request-only-in-netlify-sveltekit/46526, that's a netlify issue I found. I commented it out for now, in case we reuse it.

PR #2110, opened 2022-05-23
Add initial page about monitoring Self-Hosted

ArthurSens:
## Description

Adds an initial 🛹 page that briefly documents metrics collection and the most important signals that a user should keep an eye on when monitoring Gitpod Self-hosted.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/8566

## Release Notes

```release-note
NONE
```


ArthurSens:
I believe this is my first PR for the website, so everything is quite new to me :)

How do I add this new page to the sidebar menu? 🤔

corneliusludmann:
You need to add an entry here: https://github.com/gitpod-io/website/blob/3e5909cb1180f2ad898205a85fc31f97f81100ba/src/lib/contents/docs/menu.ts#L76-L84

ArthurSens:
Answering feedback from Slack:

>    * What endpoint are the metrics exposed at? Is this relevant? Or is this self-explanatory?

Good point, I tried to make this clearer in commit ___

>    * I would have the section about getting metrics out be more generic and split out from the rest. This then allows us have specific sections on monitoring workspaces (which the key questions and relevant metrics, monitoring server, etc).

Do you see this as a separate page? I'm not really sure how this separation should look like 🤔 

>  * This would mean we have: How to get metrics out, Metrics reference (or explainers), Dashboards& alerts as sections. Within the metrics reference we can have sections for each component, e.g. Workspace. This then means there is a good place to add more metrics later.

For a very basic monitoring, we only care about one single component: `ws-manager`. And the metrics are the ones mentioned at the page already. 

Metrics from other components are relevant to identifying causes and further debugging. Does it make sense to introduce a new page (in the future) that talks about "Advanced monitoring", which would go through metrics from other components more in depth? I suggested that because I believe we can ship this "basic monitoring" very quickly, and we shouldn't wait for an advanced guide to merge this(which would require all teams to collaborate 😬).

> * Do you have ideas on what the user should alert on?

That's a tricky one, I can suggest that people should create alerts around:
* Workspaces stuck in pending state
* Workspaces starting too slow
* Workspace failure rates

Thresholds will vary from user to user, but I'll try to make that clear in the page.

> * Tone of voice - could be a bit more brief and direct. The current tone of voice is more akin to how you speak, but when it writing might be better to be more brief. E.g. we don’t necessarily need the word precisely in “which means that it precisely represents real-time data.” Also, I’d avoid using the words “we use” - we should just state these things as a fact, i.e. this is what it is.

👍 

> * [For later] Should we provide an actual dashboard?

If we commit to this basic monitoring page, I believe I can create a dashboard 🙂. Not sure where to store this dashboard though 🤷

lucasvaltl:
> Do you see this as a separate page? I'm not really sure how this separation should look like 🤔

--> I'll try to propose something :)

lucasvaltl:
Actually, thinking more about this, I think the guide is good in the way it is - and agree it allows us to ship this quickly and iterate. In the future I think we should have a separate, more dry page about how to get metrics out - i.e. not a how to guide, but a reference about how data is exposed and where it is exposed. I am currently thinking about how to generally structure our docs more and have different types of documentation for different purposes, but this is besides the scope of this.

gitpod/docs/self-hosted/latest/monitoring.md
```diff
@@ -0,0 +1,89 @@
+---
+section: self-hosted/latest
+title: Monitoring Gitpod Self-Hosted
+---
+
+<script context="module">
+  export const prerender = true;
+</script>
+
```
lucasvaltl, line 9: To set expectations, it would make sense to add an explainer of what this is, what the goal is, and who it is for. E.g. "This guide is aimed at helping you set up basic monitoring of your Gitpod instance. In the end, you will have a continuous, high-level view of the health of your installation that you can monitor and alert on in order to respond to any issues more quickly."

gitpod/docs/self-hosted/latest/monitoring.md
```diff
@@ -0,0 +1,89 @@
...
+# Monitoring
+
+Once you have Gitpod self-hosted up and running, the next step is making sure it continues to run as expected. For that, you will probably want to set up a good monitoring solution that works well with the observability data that Gitpod produces.
+
+> **Note** All metrics shown on this page are experimental and might change in the future.
+
+## Metrics collection
+
+Several components of Gitpod expose metrics using the [Prometheus exposition format](https://prometheus.io/docs/concepts/data_model/), but for this guide, we'll focus on the most important one that makes sure that Workspaces are starting and running reasonably.
```
lucasvaltl, line 18: todo here - link to architecture diagram of the reference architecture once we have that live in the docs

gitpod/docs/self-hosted/latest/monitoring.md
```diff
@@ -0,0 +1,89 @@
...
+
+### How many workspaces are currently running
+
+To discover how many workspaces are currently running, use the [PromQL query](https://prometheus.io/docs/prometheus/latest/querying/basics/) below:
+
+```promql
+sum(gitpod_ws_manager_workspace_phase_total{phase="RUNNING"}) by (type)
+```
+
+`gitpod_ws_manager_workspace_phase_total` is a [Gauge](https://prometheus.io/docs/concepts/metric_types/#gauge). Although not suitable for alerting, with this query, you're able to tell how many workspaces, prebuilds and imagebuilds are running. This can be good information to tell how saturated your Gitpod instance is.
```
lucasvaltl, line 41: Why is this not suitable for alerting? :) Might be worth explaining this in a bit more detail.

gitpod/docs/self-hosted/latest/monitoring.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Dashboards and Alerts
+
+To have all that data available and presented in a friendly way, it is recommend building [Grafana Dashboards](https://grafana.com/grafana/dashboards/) with the most important metrics (the ones shown in this guide).
+
+Alerting can be done with [Prometheus itself](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/). If you are using the Prometheus-Operator as we recommended, you can also use the [PrometheusRule](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#prometheusrulespec) CRD to simplify alerting configuration. The [Alertmanager](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#alertmanagerspec) CRD can be used to configure alert routing to different popular platform such as [PagerDuty](https://www.pagerduty.com/) or [Slack](https://slack.com/).
```
lucasvaltl, line 89: Might be worth moving this alerting part to the top of the document (just before ## What you should keep an eye on) - we are talking about alerts above, and it seems weird to talk about how to do alerting after talking about what to alert on.

lucasvaltl (approved):
Lgtm, pending the suggestions I made :)

lucasvaltl:
Can we add a "What to do if you find an issue" section at the end in order to make this a complete loop? From there we can link to this page: https://www.gitpod.io/docs/self-hosted/latest/troubleshooting

gitpod/docs/self-hosted/latest/monitoring.md
ArthurSens: Do you mean expanding the first paragraph? 

Looking at other pages from documentation, I don't see this format being used elsewhere 😅

gitpod/docs/self-hosted/latest/monitoring.md
ArthurSens: Oof, I cannot think of a good way to explain this without teaching all the boring SRE principles 🤔. I'll put it here, please let me know if you think it should go to the docs 😅 


There are usually 4 golden signals that people should look out for when doing monitoring: _latency, traffic, errors, and saturation._ To monitor different things, it is just a matter of finding the right representation of those signals for different domains. 

For this page, I tried to translate the signals to the domain "Gitpod workspaces":
* Latency - "Workspaces are starting in a reasonable time frame"
* Traffic - "Workspaces are starting"
* Errors - "Running workspaces don't stop unexpectedly"
* Saturation - "How many workspaces are currently running"

The 3 first ones clearly represent symptoms, i.e. good alerts, but Saturation doesn't necessarily mean a bad symptom. Saturation is about measuring how much system resources a given service has. Up to a certain point, most services will experience performance degradation but such degradation is already covered by the other 3 signals. Saturation will be used mostly to understand capacity and can help define monitoring objectives and targets.

gitpod/docs/self-hosted/latest/monitoring.md
ArthurSens: Not sure if I understood correctly the urgency on this one.

The lack of a diagram should block the merge?

gitpod/docs/self-hosted/latest/monitoring.md
lucasvaltl: Yes I mean expanding the first paragraph or adding one before this. Could also make this a callout box. We do not do that yet on other pages, but we should start :)

ArthurSens:
I've already made the changes as you were giving feedback 🙂

PR #2207, opened 2022-06-10
Document `gp top` command

andreafalzetti:
## Related Issue(s)

Related https://github.com/gitpod-io/gitpod/issues/9521

## How to test
Adding documentation for the new gitpod-cli command: `$ gp top` 

Requires https://github.com/gitpod-io/gitpod/pull/10570 to be merged and deployed first.

## Release Notes

```release-note
NONE
```

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2207"></a>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2207"></a>

### DIFF
--- a/gitpod/docs/command-line-interface.md
+++ b/gitpod/docs/command-line-interface.md
@@ -31,6 +31,7 @@ Available Commands:
   sync-await          Awaits an event triggered using gp sync-done
   sync-done           Notifies the corresponding gp sync-await calls that this event has happened
   tasks               Interact with workspace tasks
+  top                 Display workspace resource (CPU and memory usage)
   url                 Prints the URL of this workspace
   version             Prints the version of the CLI
 
@@ -161,3 +162,7 @@ Outputs a table-formatted list of ports along with their status, URL, name and d
 ```sh
 gp ports list
 ```
+
+### top
+
+Displays the used and available workspace CPU and memory.


gitpod/docs/command-line-interface.md
```diff
@@ -148,3 +149,7 @@ Alternatively, specify the `Terminal ID` that you can see with `gp tasks list`:
 ```sh
 gp tasks attach <id>
 ```
+
+### top
+
+Displays the used and available workspace CPU and memory.
```
andreafalzetti, line 155: It's very short, I feel we could add more, especially about clarifying that these resources are the ones that the workspace instance is using but the limits are not necessarily _guaranteed_

andreafalzetti: If anyone is inspired or has suggestions, please shout :)

loujaybee (approved):

filiptronicek (commented):
I like the description, but perhaps we should elaborate in the section for it.

Let's mayhaps add why something about milicores in the CPU metric - I have no idea what the unit is and why we use it (I'm just guessing that's how the standard CPU consumption percentage is calculated, but the fact that I am not sure means I would document it!).

I would also add or link what `MiB` means [memory usage], and that it is not the same as `MB`, although similar [[1](https://www.techtarget.com/searchstorage/definition/mebibyte-MiB)].

This is a really cool feature and big thanks @andreafalzetti for implementing it!

andreafalzetti:
Thanks a lot @filiptronicek - that's what I needed.

Honestly, the implementation was very simple because supervisor provides already all of it! 😄 

I will give another round of thought for the description and update the PR soon. 🙏

gitpod/docs/command-line-interface.md
akosyakov: maybe we start with it and assume that it is guaranteed?

andreafalzetti:
sounds good to me, @akosyakov 👍

PR #2228, opened 2022-06-13
Fix customers page link

loujaybee:
The "view customer stories" on [gitpod.io/customers](https://gitpod.io/customers) currently links to the same page (and nothing happens when you click). 

Is this supposed to go to the anchor, instead? 

Not 100% sure if this is the right fix, as not sure where else this component is used. 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2228"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2228"></a>


chrifro:
Sorry, I'm not sure I understand the required change.

Are you talking about this component?
Bildschirmfoto 2022-06-13 um 15 44 03

There the "view customer story" button should link directly to the customer stories. But that's already the current behavior. I didn't notice any change of behavior in the preview 🤔

nisarhassan12:
@chrifro Sorry my bad I mistakenly posted the wrong screenshot I meant the Used by section  below is the correct screenshot
image

chrifro:
Ah okay, now it makes sense. Thanks @nisarhassan12 !

I'd say let's just remove the link from underneath the logos on /customers. Then component can stay as it is on all other pages

PR #2279, opened 2022-06-21
updates related to the community license page. 

nisarhassan12:


phimae:
@nisarhassan12 just reviewed the changes, wondering why you did this https://github.com/gitpod-io/website/pull/2279/commits/fd8c6b06cf25a04efbfcdb3c3ad38b8edf0f1901 ?

Our goal here is to identify the valuable leads, e.g. with high dev count and also receive emails for these. For the ones with 1-10 devs, we do not want to be distracted. 

see the task:

> Restrict emails (not the lead transfer to Google sheet) being sent to [contact@gitpod.io](mailto:contact@gitpod.io) or [sales@gitpod.io](mailto:sales@gitpod.io) for license requests of 1-10 devs as it clogs up our inboxes through the flood of requests

Trying to make it clearer:

- **All** leads (no matter 1-10 or +1000) that request a license either on the new form or on /sales should go to the Feedback Google sheet
- Leads with 1-10 should **not** trigger an email to contact@ but still create an entry in the feedback sheet (otherwise they do not receive their license)
- As a consequence: Leads with more than 10 devs should trigger **both** an email AND an entry in the feedback sheet

phimae (approved):
lgtm

PR #2302, opened 2022-06-24
Experiment: IDE toggle

loujaybee:
In a few places in the docs we have separate sections for each IDE, this introduces a toggle to showcase the different IDEs. 

|Before|After|
|-|-|
|image|image|

**Considerations:**  
- Headings are no longer seen in the overview (screenshots below)
- Formatting has to be HTML, not MD (So I guess we lose things like link validation?)
- Do we separate JetBrains IDEs down? Or always one section for JetBrains? 🤔 
- I have removed the toggle, as unlike on cloud, this toggle would likely always be showing different options intentionally, and not for abstracting away details. 

|Before|After|
|-|-|
|image|image|

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2302"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2302"></a>

### DIFF
--- a/src/lib/components/docs/ide-toggle.svelte
+++ b/src/lib/components/docs/ide-toggle.svelte
@@ -49,10 +49,14 @@
 <div class="my-8 mt-0">
   <header>
     <nav>
-      <ul class="flex flex-wrap !pl-0 !mb-0">
+      <ul class="flex flex-wrap !pl-0 !mb-0" role="tablist">
         {#each items as item}
           {#if Object.keys($$slots).includes(item.slotName)}
-            <li class="!before:hidden">
+            <li
+              class="!before:hidden"
+              role="tab"
+              aria-selected={item.value === activeValue}
+            >
               <span
                 tabindex="0"
                 class="rounded-t-2xl cursor-pointer px-4 py-2 hidden md:block {activeValue ===
@@ -77,13 +81,13 @@
   </header>
   {#if $$slots.vscodebrowser}
     {#if activeValue === 1}
-      <article class="box bg-white dark:bg-card">
+      <article class="box bg-white dark:bg-card" role="tabpanel">
         <slot name="vscodebrowser" />
       </article>
     {/if}
     {#if $$slots.vscodedesktop}
       {#if activeValue === 2}
-        <article class="box bg-white dark:bg-card">
+        <article class="box bg-white dark:bg-card" role="tabpanel">
           <slot name="vscodedesktop" />
         </article>
       {/if}
@@ -91,7 +95,7 @@
   {/if}
   {#if $$slots.jetbrains}
     {#if activeValue === 3}
-      <article class="box bg-white dark:bg-card">
+      <article class="box bg-white dark:bg-card" role="tabpanel">
         <slot name="jetbrains" />
       </article>
     {/if}
--- a/src/lib/components/docs/ide-toggle.svelte
+++ b/src/lib/components/docs/ide-toggle.svelte
@@ -79,6 +79,14 @@
           ].focus();
         }
         break;
+      case "Numpad7":
+        e.preventDefault();
+        activeValue = switchableIndexes[0];
+        break;
+      case "Numpad1":
+        e.preventDefault();
+        activeValue = switchableIndexes[switchableIndexes.length - 1];
+        break;
     }
   };
 
--- a/src/lib/components/docs/ide-toggle.svelte
+++ b/src/lib/components/docs/ide-toggle.svelte
@@ -139,4 +139,11 @@
       </article>
     {/if}
   {/if}
+  {#if $$slots.commandline}
+    {#if activeValue === 4}
+      <article class="box bg-white dark:bg-card" role="tabpanel">
+        <slot name="commandline" />
+      </article>
+    {/if}
+  {/if}
 </div>
--- a/src/lib/components/docs/ide-toggle.svelte
+++ b/src/lib/components/docs/ide-toggle.svelte
@@ -33,6 +33,40 @@
 
   const clickHandler = (tabValue: number) => () => (activeValue = tabValue);
 
+  const switchableIndexes = items
+    .filter((item) => Object.keys($$slots).includes(item.slotName))
+    .map((item) => item.value);
+
+  const switchHandler = (e: KeyboardEvent) => {
+    const currentIndex = switchableIndexes.indexOf(activeValue);
+    //@ts-ignore
+    const siblings = e.target.parentElement.children;
+    switch (e.code) {
+      case "ArrowRight":
+        {
+          e.preventDefault();
+          const willOverflow = currentIndex === switchableIndexes.length - 1;
+          activeValue = willOverflow
+            ? switchableIndexes[0]
+            : switchableIndexes[currentIndex + 1];
+          siblings[willOverflow ? 0 : currentIndex + 1].focus();
+        }
+        break;
+      case "ArrowLeft":
+        {
+          e.preventDefault();
+          const willOverflow = currentIndex === 0;
+          activeValue = willOverflow
+            ? switchableIndexes[switchableIndexes.length - 1]
+            : switchableIndexes[currentIndex - 1];
+          siblings[
+            willOverflow ? switchableIndexes.length - 1 : currentIndex - 1
+          ].focus();
+        }
+        break;
+    }
+  };
+
   export let id = "ide-toggle";
 </script>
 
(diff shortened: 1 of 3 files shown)


src/lib/components/docs/ide-toggle.svelte
```diff
@@ -0,0 +1,108 @@
+<script lang="ts">
```
loujaybee, line 1: Mostly copy/pasta 🍝 from [cloud platform toggle](https://github.com/gitpod-io/website/blob/main/src/lib/components/docs/cloud-platform-toggle.svelte).

src/lib/components/docs/ide-toggle.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+  export let items: comparisonItem[] = [
+    {
+      mobileTitle: "JetBrains",
+      title: "JetBrains",
+      value: 1,
+      slotName: "jetbrains",
+    },
+    {
+      mobileTitle: "VS Code Browser",
```
Cahllagerfeld, line 14: |IDE|Architecture|
|-|-|
|grafik|grafik|

Could we potentially have any abbreviations to display on mobile breakpoints? For Cloud Platforms we're using `GCP` instead of the full-blown `Google Cloud Platform` for example

Cahllagerfeld (commented):
Just a small comment. I pushed a small fix for dark-mode (not caused by you :D)

src/lib/components/docs/ide-toggle.svelte
filiptronicek: @Cahllagerfeld I'm not sure if users are familiar with calling JetBrains JB and Visual Studio Code VS Code. I'm worried that abbreviations will cause confusion :/

felladrin:
How about showing VS Code Browser as the first (initial) tab, as it's the only one that currently has an image?

So the order of tabs would be: `VS Code Browser, VS Code Desktop, JetBrains`.

src/lib/components/docs/ide-toggle.svelte
loujaybee: Hrm, it's challenging...

We'd have to break some conventions that we have throughout the docs to make it work, e.g...

Do we have a minimum mobile breakpoint width that I can optimise for, btw? 

image

loujaybee:
> How about showing VS Code Browser as the first (initial) tab, as it's the only one that currently has an image?

I do agree, however it looks like the way the code is structured, the order is hard-coded. I was intending to keep code similar / same to the [cloud platform toggle](https://github.com/gitpod-io/website/blob/main/src/lib/components/docs/cloud-platform-toggle.svelte). I could give a swing, but wouldn't want to step on @Cahllagerfeld / @nisarhassan12's toes too much... we could add as an optimisation / follow up and apply to both toggle components?

filiptronicek:
> How about showing VS Code Browser as the first (initial) tab, as it's the only one that currently has an image?

@loujaybee I don't think it's necessary to make it dynamic as of now, but I would see value in just hard-coding a different order of these tabs to make VS Code Browser the default, since it has the most users out of all - no switching for most users. We could then order them by approximate usage so that users reading from left to right are most likely to find their IDE rather sooner than later - this being a very small nitpick though, saving a couple of hounded milliseconds at best 😆.

Also had an idea for the future - a shared cookie like the now present `gitpod-user` (which includes info about whether the user is logged in or not) but for their preferred IDE, so that by default, they would have their IDE of choice always selected - with the rest being in a dropdown to reduce the visual space needed. 

Would be down to take this one in the future, would be cool to work on 🙂.

filiptronicek (commented):
Maybe we should give the tab component another spin (for both this one and the cloud platform toggle). There are some accessibility concerns I have if we were to introduce it (acknowledging this is an experiment, but I would love to see it soon in the docs 🙂):
1. We don't have any ARIA accessible attributes on our component, so screen readers kind of just read the headings (without understanding they are navigation elements) and then read only the current active tab content. We should make the titles be inside a `<nav>` with `<a>`s as buttons. A bit like [here](https://github.com/argyleink/gui-challenges/blob/5f293302cc50d9151db88824b85e87a8b903abfa/tabs/app/index.html#L28-L40).
2. The tabs aren't navigable via the keyboard, so for anyone who wants to utilize it / has it as their only input option they are out of luck.

There are some more optimizations on my mind that could be made, but I would be happy to see some more love put into it, if we want to expand its use throughout the docs. 

I recognize this is not IDE-toggle specific, sorry! I'll try to take a look at it from that aspect again too.

Cahllagerfeld:
Nice comment @filiptronicek 🙏 When the idea for the component came up, it was a used-once thing for Self-Hosted Reference Architecture and meant as a Skateboard to enable them. When seeing now how frequently it should get used, it definitely needs to get enhanced and polished. The issue that component has is in the nature of markdown: As these pages are static and rendered at compile-time, our Page doesn't know which Tab is selected - ergo the "on-this-page"-navigation doesn't know which tab is selected and can't react to headlines that are inside this tab. That's why we need to avoid having headlines inside the tabs.

src/lib/components/docs/ide-toggle.svelte
```diff
@@ -0,0 +1,93 @@
...
+            class="rounded-t-2xl cursor-pointer px-4 py-2 md:hidden block {activeValue ===
+            item.value
+              ? 'bg-white dark:bg-card'
+              : 'bg-sand-dark dark:bg-light-black'} transition-all duration-200"
+            on:click={clickHandler(item.value)}>{item.mobileTitle}</span
+          >
+        </li>
+      {/if}
+    {/each}
+  </ul>
```
filiptronicek, line 71: Added keyboard navigation, please see a short Loom I made about this: https://www.loom.com/share/77bc952fc7cd4114a72ac401e54fa74b.

```suggestion
  <header>
    <nav>
      <ul class="flex flex-wrap !pl-0 !mb-0">
        {#each items as item}
          {#if Object.keys($$slots).includes(item.slotName)}
            <li class="!before:hidden">
              <span
                tabindex="0"
                class="rounded-t-2xl cursor-pointer px-4 py-2 hidden md:block {activeValue ===
                item.value
                  ? 'bg-white dark:bg-card'
                  : 'bg-sand-dark dark:bg-light-black'} transition-all duration-200"
                on:click={clickHandler(item.value)}
                on:focus={clickHandler(item.value)}>{item.title}</span
              >
              <span
                class="rounded-t-2xl cursor-pointer px-4 py-2 md:hidden block {activeValue ===
                item.value
                  ? 'bg-white dark:bg-card'
                  : 'bg-sand-dark dark:bg-light-black'} transition-all duration-200"
                on:click={clickHandler(item.value)}>{item.mobileTitle}</span
              >
            </li>
          {/if}
        {/each}
      </ul>
    </nav>
  </header>
```

src/lib/components/docs/ide-toggle.svelte
```diff
@@ -0,0 +1,93 @@
...
+      <div class="box bg-white dark:bg-card">
+        <slot name="vscodebrowser" />
+      </div>
+    {/if}
+  {/if}
+  {#if $$slots.vscodedesktop}
+    {#if activeValue === 3}
+      <div class="box bg-white dark:bg-card">
+        <slot name="vscodedesktop" />
+      </div>
```
filiptronicek, line 90: We could use `<article>` to indicate every tab is its own "thing". 

```suggestion
      <article class="box bg-white dark:bg-card">
        <slot name="jetbrains" />
      </article>
    {/if}
  {/if}
  {#if $$slots.vscodebrowser}
    {#if activeValue === 2}
      <article class="box bg-white dark:bg-card">
        <slot name="vscodebrowser" />
      </article>
    {/if}
  {/if}
  {#if $$slots.vscodedesktop}
    {#if activeValue === 3}
      <article class="box bg-white dark:bg-card">
        <slot name="vscodedesktop" />
      </article>
```

filiptronicek:
Added some ARIA attributes :) 

image

filiptronicek:
I believe now this PR is ready to go: if I didn't make a mistake with the implementation (I doubt it) this component should be 100% ARIA-compliant. It implements all of the necessary attributes for `tab`, `tablist` and `tabpanel` with all of the [keyboard interactions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role#keyboard_interactions). Tomorrow I will post a Loom here detailing the changes.

This was quite fun 😄. I would love to see what you think, @loujaybee, @nisarhassan12, @Cahllagerfeld!

I suppose we should afterwards also update the cloud platform toggle to not have so many differences between two same-looking components 😆.

filiptronicek:
If anyone has any further thoughts, feel free to leave them here. I'm merging this now and have created https://github.com/gitpod-io/website/pull/2408 to update the Cloud Platform toggle as well.

PR #2337, opened 2022-06-29
Restructure Self-Hosted Docs Section around Types of Documentation

lucasvaltl:
## Description
This is an implementation of https://github.com/gitpod-io/gitpod/issues/10465 which itself was discussed in this (internal) [RFC](https://www.notion.so/gitpod/Proposal-Day-2-Self-Hosted-Docs-7708eea39c5e4a1fb05b04ccd99aa797). 

Specific changes:
- Restructured docs along the type of documentation
- Added an alpha tag to Reference Architectures to clarify the difference between these and the cluster set up guides

This is more an _intermediary_ step - I am trying to keep the PRs small and targeted. I have already laid out the following follow up work to further improve on this: 
 - [ ] Restructure repo: We should restructure the self-hosted docs in the repo to have one folder per subsection. This will break a lot of links though - I don't know how to best do this from an execution perspective. 
 - [ ] UX improvements: This change makes the docs look empty. We should have the sub-menus default to open for the self-hosted section to facilitate jumping between them and facilitate discovery (at least until we have too many menu items that it becomes untenable). Should probably better indicate in the UI which menu items are a subsection to facilitate discovery of docs pages.
 - [ ] Rename Self-Hosted Section into `Administer Gitpod`
 - [ ] Rename current getting started guide into installation guide. Subsequently rename local preview into getting started. Move getting started (formerly local preview) to the top Self-Hosted hierarchy level 
 - [ ] Should move TLS guide one level up the hierarchy (under installation Guides)
 - [ ] Split releases into release policy and upgrade guides (maybe move the latter to the operational guides section?
 - [ ] Need to improve the trouble shooting section with info on how to reach out to support / community. 
 - [ ] Add background section for (future) topics like Architecture Basics, Disaster Recovery, Update strategies, and TLS certificates basics - once we have the first one of these.
 - [ ] Split How to update Gitpod docs into two - update guide and update strategies. Move update strategies into the background subsection. 
- [ ] Move reference architectures up one level once we have them in a V1 state and can remove the cluster set up guides

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/10465

## How to test
Open netlify preview, navigate to docs > self-hosted

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2337"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2337"></a>


gitpod/docs/self-hosted/latest/reference.md
```diff
@@ -0,0 +1,17 @@
...
+  export const prerender = true;
+</script>
+
+# Gitpod Self-Hosted Reference
+
+Reference information for Gitpod Self-Hosted:
+
+- [Required Components](./required-components) - Information on which components Gitpod requires and with which versions of them Gitpod is compatible
+- [Releases](./releases) - Information about how Gitpod is released
+- [Telemetry](./telemetry) - Information about the telemetry that is gathered
```
Siddhant-K-code, line 17: I think this can be more clear?. Something like this:

```suggestion
- [Gitpod Self Hosted Releases](./releases) - Information about how Gitpod is released
- [Gitpod Telemetry](./telemetry) - Information about the telemetry that is gathered
```

corneliusludmann: Given that the whole page is about Gitpod Self-Hosted, this looks redundant to me. :thinking:

corneliusludmann (approved):
Looks good so far.

One limitation: Navigating a level deeper (e.g. to [TLS](https://deploy-preview-2337--gitpod-kumquat.netlify.app/docs/self-hosted/latest/advanced/tls) or the reference architectures), the menu tree collapses. Does it need a `subsection` header in these files (e.g. `tls.md`) as well?

julia-leyton:
This is looking great @lucasvaltl. I really like how things are organized now and seeing fewer sections grouping relevant docs!

I noticed the same as Cornelius, when you go a level deeper the menu collapses. Example:

Bildschirmfoto 2022-07-04 um 21 12 52
 

On the Cluster Setup section, would it make sense to add the same disclaimer shown on the installation Page? 
`Help with setting up your Kubernetes cluster. This section will soon be replaced by our [Reference Architectures]`

Bildschirmfoto 2022-07-04 um 21 43 40

Other thing, not sure if out of scope for this PR, but I really like from the Cluster Setup section that it summarizes what you will have after installation for each cloud provider.

> The installation process takes around twenty minutes. In the end, the following resources are created

I didn't see that in the Reference Architecture section and I think it is really valuable to have a short list of all the resources created after going through the whole setup process, I hope that won't go away once that section is replaced!

julia-leyton (approved):

lucasvaltl:
Thanks both! I've added `subsections` into the headers to keep the menu expanded. Overall, I think when we have reference architectures fully rolled out we can restructure this a bit nicer. I'd like to avoid these pages (e.g. the TLS page which sits under advanced installation) that link to other pages and instead have everything at the same level. This should also help with discoverability.

PR #2374, opened 2022-07-06
Adjust Code Fence

Cahllagerfeld:
Polishes the styling of the compoenet


Cahllagerfeld (commented):
Hey @chrifro, 
I personally feel like, that 12px for copy is a bit too low, wdyt?

chrifro:
I think it works well. It is good to have a clear hierarchy of elements. If everything is the same weight, the viewer has struggles to find a clear reading path. 

One small outstanding thing:
- the color of the icon and the copy text on hover are not changing simultaneously. Currently, the font color of the text changes first followed by the icon.

Otherwise, it looks 🔥. Thanks for your initiative to implement Code Fence 🚀

PR #2390, opened 2022-07-07
Update change entry date

gtsiolis:
## Description

@Cahllagerfeld I'm a bit confused from the existing changelog process, but opened this PR to update the **changelog entry date** in case this is fine according to the changelog process. Otherwise, feel free to close this.

Cc @loujaybee @atduarte @jldec because of the [relevant discussion](https://gitpod.slack.com/archives/C022HAT0H5Z/p1656919053988129) (internal).


Cahllagerfeld (commented):
Hey @gtsiolis, from my feeling I'd keep the date to be consistent with the 2 week cadence. In case we decide to publish more frequently and more agile, i think we can be quite flexible with dates. Wdyt?

gtsiolis:
@Cahllagerfeld I'm fine with any option! Let's close this then. 📕 

[FWIW](https://en.wiktionary.org/wiki/FWIW), it feels like like we're tackling ourselves with the fictitious release date on the changelog entry, especially if we'd like to measure and improve this task or process, but let's keep it as is for now.

Looping in @MichaelAring for visibility. 🏓

Cross-linking also a [relevant discussion](https://gitpod.slack.com/archives/C022HAT0H5Z/p1654856065505649) (internal).

PR #2392, opened 2022-07-08
feature: move education customers to for education

jacobparis:
## Description

This PR
- Removes education customers from /customers, so that page can focus on enterprise
- Adds permanent redirect from `/customers/bruno-rocha` and `/customers/4geeks` to their `/for/education` counterparts so inbound links don't break
- Adds education customers to the `/for/education` landing page

## How to test

Navigate to `/customers`. The customer stories should look like this and not contain any education stories

image

Navigate to `/customers/4geeks` by manually entering it in the URL bar. It should redirect you to `/for/education/4geeks`
Navigate to `/customers/bruno-rocha` by manually entering it in the URL bar. It should redirect you to `/for/education/bruno-rocha`

Navigate to `/for/education`. You should now see customer stories near the bottom of the page

image

## Release Notes

```release-note
```


Cahllagerfeld (commented):
It got a bit bigger than I expected, the one thing I wanted to achieve was use Netlify-Redirects instead of Svelte-Redirects as they are cheaper as they are not rendered or bundled etc...

I had to update the stories component for this, as it wasn't implemented yet for the use with /for/education

Issue #2357, opened 2022-07-04
Add gitpod-io/openvscode-server to changelog


loujaybee:
### Describe your feature request

We should add openvscodeserver updates to our website as well

### User impact

Showcase VS Code updates

Cahllagerfeld:
Hey @loujaybee, is this PR still relevant when having the "new" changelog in place now?

Issue #2341, opened 2022-06-30
Add `gitpod-io/workspace-images` to changelog


Cahllagerfeld:
### Describe your feature request

We should add the Workspace-Image updates to our website as well

### User impact

Being transparent about Workspace changes is mandatory to see if your project will work or break

Siddhant-K-code:
also related: https://github.com/gitpod-io/dazzle/pull/56

PR #2405, opened 2022-07-11
feat(changelog): add endpoints for in product changelog

mustard-mh:
## Description

Add endpoints for in product changelog:
- Date of latest CHANGELOG
- Query raw markdown of CHANGELOG

## Related Issue(s)

Fixes #

## How to test

**Endpoint for latest changelog data**
- /changelog/latest

**Endpoints for raw markdown content**
- /changelog/raw-markdown
- /changelog/raw-markdown?releaseId=2022-06-29
- /changelog/raw-markdown?releaseId=2021-04

## Release Notes

```release-note
NONE
```

## Documentation

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2405"></a>


loujaybee:
Some questions I think we need to answer before we merge this: 

1. **Load/Infra** - Is the current website infrastructure appropriate / prepared to receive these requests? What is the expected load? Will there be an impact on the website itself? Is the website infrastructure sufficient to expose this information? 
2. **Contract** - What are the limitations of using the markdown in this repo as a contract? What could be the impact of refactoring or restructuring in this website on the production gitpod app? Is there a possibility for changes in this repo to affect the production app? 
3. **Documentation/expecatations** - What, where, how do we want to document this behaviour if we add it to the website? How do we ensure that future maintainers of the website repo are aware of the coupling / behaviour here and the impact of updating this functionality?

jeanp413:
> Is there a possibility for changes in this repo to affect the production app?

Do you mean the gitpod server, or the vscode extensions? If the former then none, if it's the latter and if the endpoints break then we just won't show the release notes, nothing critical

mikenikles:
Julian and I discussed this PR and we agreed the website repo can't be the source of truth for changelog content.

There are requests to include additional repositories in the changelog (#2341, #2357), which is a strong indication that the `website` repo's job is to pull (rather than push) changelog content.

Julian is going to work on a prototype tomorrow to explore an architecture where individual `gitpod-io/*` repositories host their own changelog content as Markdown files. The website will fetch that content at build time and generate the `/changelog` page(s).

We will provide further instructions once we evaluated the prototype.

jeanp413:
I don't think it's a good Idea to have a changelog per repo, at least for vscode extensions, they have their how release cycle, I could release 5 new versions in a week, so how would you keep track of all the new versions between gitpod releases?, unless we sync the release cycles of vscode extension with gitpod releases but that's a bad idea too then we wouldn't be able to push fixes/features as fast as we do now.

Even if this gets implemented and website fetch changelogs from other repos, we still want to show the aggregated changelog from all gitpod components not the changelog of just the vscode extension (as I understand it). 
So website will be the source of truth of thel agregated changelog and this PR still makes sense as we need the generated markdown to exposed via an endpoint

mikenikles:
As of today, I'm aware of changelog content from `gitpod-io/gitpod` and the two repos listed in my previous comment. Are there additional repositories we maintain for VS Code extensions that also have changelog entries we want to display?

Lou's point _2. Contract_ is the main motivation for hosting changelog Markdown files outside the `gitpod-io/website` repo. The current PR's chained `.replace()` calls are fragile and I expect them to break within weeks as we continue to improve on the `/changelog` page(s).

Let me clarify what I meant by "the website repo can't be the source of truth for changelog content": I referred to the actual Markdown files. Julian and I imagine a `changelog/*.md` file structure in the `gitpod-io/gitpod` and any other repo which wants to display its changelog. The website manages a list of repos to pull that content from and generates the `/changelog/*` pages at build time. Over the last year, I've seen that having docs in the `website` repo was a mistake, hence I don't want to repeat that with the changelog content.

Separate from the actual Markdown content, I understand your ask for API endpoints where we can fetch the consolidated changelog that contains everything across the company (i.e. all repositories that provide changelog content). This can live in the `website` repo similarly to what this PR suggests, but we have to find a data structure that avoids chained `.replace()` calls.

jeanp413:
>  Are there additional repositories we maintain for VS Code extensions that also have changelog entries we want to display?

For IDE team we have [gitpod-vscode-desktop](https://github.com/gitpod-io/gitpod-vscode-desktop) and then `gitpod-web` and `gitpod-remote` extension but those are inside openvscode-server/gp-code branch (we'll move them to it's own repo in the future)

> Over the last year, I've seen that having docs in the website repo was a mistake, hence I don't want to repeat that with the changelog content.

Ah I see what you mean, doing something like [vscode-docs](https://github.com/microsoft/vscode-docs) that repo only contains the markdown, images for docs/changelogs/blogpost decoupled from the website itself

>  but we have to find a data structure that avoids chained .replace() calls.

Agree, ideally they would be just plain markdown, it's just like this for now so we have a working prototype, we could return the current file content as it is, and move the transformation with `replace()` calls into the vscode extension itself, it's fine if it breaks as we can keep iterating on it quickly as the website migrates to the new approach

Cahllagerfeld:
@mikenikles We might have an issue with the Vite-Version here. We're using `2.7.6`. The feature to import raw files was introduced with [2.8.0](https://github.com/vitejs/vite/blob/v2.8.0/packages/vite/CHANGELOG.md#280-beta0-2022-01-05). Because of that we can only import the file with the `render`-function - similar to what we do in `hooks.ts`. I'm not sure if there's a workaround for importing the raw markdown instead of the rendered html 🤔

jeanp413:
@Cahllagerfeld took a look and I think we can [use this as a workaround](https://github.com/vitejs/vite/issues/5304#issuecomment-944243698) could test it? :pray: 
```
const modules = import.meta.glob('./assets/svgIcon/*.svg')
const rawModules = {}
for (const path in modules) {
  rawModules[path] = () => import(`${path}?raw`)
}
```

Cahllagerfeld:
Hey @jeanp413
I've literally just dropped off the call with @mikenikles. Mike is going to have a look into it today, so if everything goes fine, I am able to pick this up again on Monday

Cahllagerfeld:
I've spent some time in it today, and the first impression looks quite promising. The implementation is not final yet, but the new Vite Feature works. See [Snapshot](https://gitpod.io#snapshot/a96557ba-1511-4d68-8028-ab3124c5f60e)

Cahllagerfeld:
I've spent some time in it today, and the first impression looks quite promising. The implementation is not final yet, but the new Vite Feature works. See [Snapshot](https://gitpod.io#snapshot/a96557ba-1511-4d68-8028-ab3124c5f60e)

Cahllagerfeld:
Based on #2439, I've created the following [snapshot](https://gitpod.io#snapshot/602ffd74-c980-4638-a3e1-200e8b5cde6f) that contains the implementation. @mikenikles I think the question is the order in which we want to merge/push things

Cahllagerfeld:
Hey @jeanp413, I've upgraded the Sveltekit version today, to enable and unlock the needed features. I also pushed the discussed changes to this PR. From my side you can start running your first tests again this Preview Deployment, to see if everything works as expected

src/routes/changelog/latest.ts
```diff
@@ -0,0 +1,10 @@
+export const get: import("@sveltejs/kit").RequestHandler = async ({
+  locals,
+}) => {
+  const { date } = locals.changelogEntries[0];
+  return {
+    body: {
+      date,
```
jeanp413, line 7: @mustard-mh this should return only the date YYYY-MM-DD right now it returns second, miliseconds etc

src/routes/changelog/raw-markdown.ts
```diff
@@ -0,0 +1,33 @@
+import type { ChangelogEntry as ChangelogEntryType } from "$lib/types/changelog-entry.type";
+import path from "path";
+
+export const get: import("@sveltejs/kit").RequestHandler = async ({
+  url,
+  locals,
+}) => {
+  const changelogEntries = Object.entries(
+    import.meta.globEager("/src/lib/contents/changelog/**.md", { as: "raw" })
```
jeanp413, line 9: maybe we should use `import.meta.glob`, I think `globEager` loads all md files at once

mustard-mh:
@jeanp413 @Cahllagerfeld Rebase and fix some bugs, it's ready for review not 🙏

jeanp413 (approved):

jeanp413:
@Cahllagerfeld waiting for your review :pray:

src/routes/changelog/raw-markdown.ts
```diff
@@ -0,0 +1,33 @@
+import { basename } from "path";
+
+export const get: import("@sveltejs/kit").RequestHandler = async ({
+  url,
+  locals,
+}) => {
+  const changelogEntries = Object.entries(
+    import.meta.glob("/src/lib/contents/changelog/**.md", { as: "raw" })
```
jeanp413, line 8:
```suggestion
    import.meta.glob("/src/lib/contents/changelog/*.md", { as: "raw" })
```

Cahllagerfeld (approved):
Let's get it merged 🚀

PR #2462, opened 2022-07-22
Remove procuction-ready mention from single cluster reference architecture

lucasvaltl:
## Description
As discussed internally, we change the way we talk about the single-cluster reference architecture. It is still intended for production use cases, and this is made clear even with the change proposed here. The badge of `production ready` may mean different things for different people, hence why we want to be a bit careful here at first. 

## How to test
Open preview, navigate to docs>self-hosted>installation guides > reference architectures > single cluster reference architecture

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2462"></a>


adrienthebo (approved):
LGTM!

PR #2468, opened 2022-07-25
docs: delete bash page

ghuntley:
## Description

I'm proposing here that the `bash` page should not exist in the first place:

* References old information about bugs that are over two years old
* Suggesting folks install a language server (authored in javascript) to debug `#!/usr/bin/env bash` is odd.
* Uses old screenshot.
* Gitpod ships with bash as the default shell. I don't see why a page is needed because _there's nothing special about Gitpod, it's a general-purpose code execution platform_. Having the page implies _bash needs to be configured on Gitpod_?

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2468"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2468"></a>


Cahllagerfeld (commented):
I've fixed the preview deployment. 
We should add a Netlify-Redirect so external links pointing to the bash page don't end up with 404. Where should we redirect this URL to @ghuntley wdyt?

ghuntley:
> Where should we redirect this URL to @ghuntley wdyt?

/docs

Siddhant-K-code (approved):

Issue #2454, opened 2022-07-21
Update Changelog Template


Cahllagerfeld:
- Roadmap Updates
- Template for Other Components
- `**Latest Release:** You can enable Latest Release in [your preferences](https://gitpod.io/preferences) and try the latest updates as soon as they are available.`

loujaybee:
@Cahllagerfeld all good if I pick this one up, and work on it with @filiptronicek ?

PR #2473, opened 2022-07-26
Update changelog template (latest + roadmap section)

loujaybee:
Proposes some updates to the changelog template: 

1. Add **"latest"** section for early features (with instructions on how to enable, etc). 
2. Add **roadmap** section - This has been added for lots of past changelog entries but was not yet in the template. 
3. Add **deprecations** section - For upcoming features / API's to be removed. 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2473"></a>

## Context

We recently added a product changelog which is shown inside VS Code Desktop + Browser to Gitpod users: 

- https://github.com/gitpod-io/gitpod/issues/7537

We want to make sure to highlight these latest features in the changelog by adding an explicit section which would draw attention to these early / pre-release features. 

Closes: https://github.com/gitpod-io/gitpod/issues/11622

### DIFF
--- a/src/lib/contents/changelog/_template.md
+++ b/src/lib/contents/changelog/_template.md
@@ -4,7 +4,7 @@ excerpt:
 date: {{releaseDate}}
 image: {{releaseDate}}.jpg
 alt:
-tag: //incase its a self-hosted Changelog add the value `self-hosted`, otherwise tis prop can be left empty
+tag: //in case its a self-hosted changelog add the value `self-hosted`, otherwise this prop can be left empty
 ---
 
 <script>
--- a/src/lib/contents/changelog/_template.md
+++ b/src/lib/contents/changelog/_template.md
@@ -14,6 +14,21 @@ TODO: Insert the featured changelog entry here.
 
 <p><Contributors usernames="" /></p>
 
+### Roadmap updates
+
+<div class="mt-medium">
+
+**FeatureName** - Roadmap issue: [#XXXX](https://github.com/gitpod-io/gitpod/issues/XXXX) <Badge text="beta" variant="orange" class="ml-1.5" />
+
+</div>
+
+### Latest 
+
+The following are beta Gitpod features not currently in [general availability](https://www.gitpod.io/docs/references/gitpod-releases). If something catches your eye, why not try it out, and let us know what you think in the [community]([url](https://community.gitpod.io/)), or if needed, [raise an issue]([url](https://github.com/gitpod-io/gitpod)).
+
+**Please note:** Some features might require a specific IDE or editor version, or certain configuration steps to be enabled. 
+
+
 ### Fixes and improvements
 
 {{fixesAndImprovements}}
--- a/src/lib/contents/changelog/_template.md
+++ b/src/lib/contents/changelog/_template.md
@@ -4,7 +4,7 @@ excerpt:
 date: {{releaseDate}}
 image: {{releaseDate}}.jpg
 alt:
-tag:
+tag: //incase its a self-hosted Changelog add the value `self-hosted`, otherwise tis prop can be left empty
 ---
 
 <script>
--- a/src/lib/contents/changelog/_template.md
+++ b/src/lib/contents/changelog/_template.md
@@ -28,6 +28,11 @@ The following are beta Gitpod features not currently in [general availability](h
 
 **Please note:** Some features might require a specific IDE or editor version, or certain configuration steps to be enabled. 
 
+### Deprecations
+
+Please ensure to update any necessary Gitpod enabled repositories, configurations or installations in accordance with the following deprecations:
+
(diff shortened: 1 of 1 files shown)


src/lib/contents/changelog/_template.md
```diff
@@ -14,6 +14,21 @@ TODO: Insert the featured changelog entry here.
...
 
+### Roadmap updates
+
+<div class="mt-medium">
+
+**FeatureName** - Roadmap issue: [#XXXX](https://github.com/gitpod-io/gitpod/issues/XXXX) <Badge text="beta" variant="orange" class="ml-1.5" />
+
+</div>
+
+### Latest
```
loujaybee, line 25: CC: @akosyakov / @jeanp413 if you have more suggestions.

loujaybee, line 25: CC: @akosyakov / @jeanp413 if you have more suggestions.

src/lib/contents/changelog/_template.md
```diff
@@ -14,6 +14,21 @@ TODO: Insert the featured changelog entry here.
 
 <p><Contributors usernames="" /></p>
 
+### Roadmap updates
```
loujaybee, line 17: CC: PM's, we've had this section for the last few changelogs, see: 

- https://www.gitpod.io/changelog/bring-your-own-workspace-ssh-keys
- https://www.gitpod.io/changelog/auto-cancel-prebuilds-on-outdated-commits

Might be good to think about how we can leverage this idea more, e.g. tie PR's and issue changes to the larger roadmap items to make them easier to digest for users, and see progress towards the roadmap items.

Cahllagerfeld:
This thing, was on my todo list for today as well 🤯 Looks like you've been faster :D

src/lib/contents/changelog/_template.md
```diff
@@ -14,6 +14,26 @@ TODO: Insert the featured changelog entry here.
...
+
+</div>
+
+### Latest 
+
+The following are beta Gitpod features not currently in [general availability](https://www.gitpod.io/docs/references/gitpod-releases). If something catches your eye, why not try it out, and let us know what you think in the [community]([url](https://community.gitpod.io/)), or if needed, [raise an issue]([url](https://github.com/gitpod-io/gitpod)).
+
+**Please note:** Some features might require a specific IDE or editor version, or certain configuration steps to be enabled. 
+
+### Deprecations
```
loujaybee, line 31: Going to use this opportunity to add a deprecations section. 

This is something I've wanted/needed for quite some time...

**Current example:** 
- Local Companion usage for VS Code Desktop (https://www.gitpod.io/docs/ides-and-editors/vscode)

**Past example:** 
- JetBrains projector support

src/lib/contents/changelog/_template.md
akosyakov: It would be good to have this point always even if we don't put anything explicit, since we have many small changes there from upstream. 

So something what VS Code team adds to each release note:
> Insiders: Want to try new features as soon as possible? You can download the nightly [Insiders](https://code.visualstudio.com/insiders) build and try the latest updates as soon as they are available.

We could have something like:
> **Latest Release:** You can enable Latest Release in [your preferences](https://gitpod.io/preferences) and try the latest updates as soon as they are available.

Actually tracked in https://github.com/gitpod-io/website/issues/2454

jeanp413: I don't think this section makes sense, the concept of changelog itself is supposed to be about the latest changes at least for me :thinking: 

> Latest Release: You can enable Latest Release in [your preferences](https://gitpod.io/preferences) and try the latest updates as soon as they are available.

That would be ide related so I'd said just put that in the corresponding ide section, if there's something to showcase

src/lib/contents/changelog/_template.md
```diff
@@ -4,6 +4,7 @@ excerpt:
 date: {{releaseDate}}
 image: {{releaseDate}}.jpg
 alt:
+tag:
```
Cahllagerfeld, line 7: According to the [Slack-Discussion](https://gitpod.slack.com/archives/C01KGM9BH54/p1659017033343739) I've added this to the frontmatter cc: @akosyakov

src/lib/contents/changelog/_template.md
akosyakov: I also think both `Roadmap` and `Latest` does not really make sense. I would prefer to have headers to be new features either they go to stable or latest.

src/lib/contents/changelog/_template.md
akosyakov: Can we have a comment here? 
Something like for self-hosted release, use `tag: self-hosted`?

loujaybee:
Will close pending further discussions.

PR #2476, opened 2022-07-26
[self hosted] Document Kubernetes privileged pod requirements

adrienthebo:
## Description

Gitpod requires the ability to create and manage PodSecurityPolicy resources and run privileged pods; this does not mesh well with organizations that default to restrictive cluster policies. This pull request adds documentation to this effect.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2476"></a>


lucasvaltl (approved):
LGTM - although this page will likely be replaced by, or merged into the reference architecture. Still thinking about how exactly to do this :)

PR #2482, opened 2022-07-27
Prefer "email" over "e-mail"

jankeromnes:
Hi there! 😇 There is just a super small fix -- I found the spelling of "e-mail" looked a bit strange in the webinar registration form, and I think we should use the simpler, more common spelling "email".

To be fair, both "e-mail" and "email" spellings are correct -- however, "email" is vastly more popular and I also find it looks nicer.

Screenshot 2022-07-27 at 09 29 32

Source: https://trends.google.com/trends/explore?q=email,e-mail

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2482"></a>


Cahllagerfeld (approved):
Nice catch! 🚀

PR #2485, opened 2022-07-27
Redirect community.gitpod.io even if no path is given

meysholdt:
## Description
Follou-up fix for https://github.com/gitpod-io/website/pull/2483: Do the redirect even if no path is given.

## Related Issue(s)
Fixes https://github.com/gitpod-io/website/pull/2483#issuecomment-1196814197

## How to test
Can't test before merge :(

## Release Notes

```release-note
NONE.
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2485"></a>

PR #2488, opened 2022-07-27
Improve redirects

meysholdt:
## Description
The redirect for blog.gitpod.io and community.gitpod.io does not work if the domains are not followed by a path.
Unfortunately, the fix from https://github.com/gitpod-io/website/pull/2485 does not resolve this.

Diving deeper into the docs unveils that redirects by default have lower precedence than static files. 
Since we want a redirect in *any* case, we need to add a trailing`!` to the rule.

See the [docs for more details](https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing).

This PR also preserves the path for bog posts during redirects and effectively reverts https://github.com/gitpod-io/website/pull/2485.

## How to test
Not possible with Netlify preview envs I'm afraid. 

## Release Notes

```release-note
NONE
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2488"></a>


_redirects
```diff
@@ -1,7 +1,5 @@
-https://blog.gitpod.io https://www.gitpod.io/blog 301
-https://blog.gitpod.io/* https://www.gitpod.io/blog 301
-https://community.gitpod.io https://www.gitpod.io/chat 301
-https://community.gitpod.io/* https://www.gitpod.io/chat 301
+https://blog.gitpod.io/* https://www.gitpod.io/blog/:splat 301!
+https://community.gitpod.io/* https://www.gitpod.io/chat 301!
```
Cahllagerfeld, line 2: Will this work, or will we run in the thing that it doesn't work for the root URL, as these are only two rules? 😅

meysholdt: I hope it will work :) 

> will we run in the thing that it doesn't work for the root URL

that's what this change aims to fix. The rules without the `/*` never worked, so this PR removes them. I belive Netlify's concept of [Shadowing](https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing) is the culprit and the 2nd pair of rules was never needed.

Cahllagerfeld (approved):
Lets merge it 🚀

PR #2518, opened 2022-08-02
[PROPOSAL] docs(configure-ide): proposal for rework

ghuntley:
## Description

We received the following feedback:

> Can you configure the _selection_ of an IDE per project? This page seems to imply that you can, yet https://www.gitpod.io/docs/references/gitpod-yml does not have any such apparent option. If it is _not_ an option, OK but please say so explicitly.

which prompted a content review cycle and spawned the following thoughts:

1. The page `/docs/ides-and-editors/configure-your-editor-ide` page should be completely deleted and replaced with configuration instructions as subheadings on each editor's page. See the attached changes for suggestions of what each subheading could look like.
2. The page `/ides-and-editors` should be updated with instructions on how to switch editors, switching editors is account level and there's no way to enforce configuration (why would someone want this?) at a team level.
3. Mentioning "per project configuration of Gitpod.yml" in the context of start tasks or doesn't make sense to me here unless it is in the context of "installation of custom software into the Gitpod environment aka bring your own custom docker image". Thoughts here is that any such reference should be 🆙  a level at `/ides-and-editors`

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2518"></a>

### DIFF
--- a/gitpod/docs/ides-and-editors/configure-your-editor-ide.md
+++ b/gitpod/docs/ides-and-editors/configure-your-editor-ide.md
@@ -9,24 +9,35 @@ title: Configure your IDE/editor
 
 # {title}
 
-Customise the IDE or editor experience for a project, or yourself.
+Which IDE/editor is used when launching workspaces is configured via your [user preferences](https://gitpod.io/preferences) and take effect upon launching new workspaces.
 
-## Per Project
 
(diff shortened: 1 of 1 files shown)


loujaybee:
Agree to just remove the page, added a PR here: 

https://github.com/gitpod-io/website/pull/2547

PR #2534, opened 2022-08-03
Includes bruteforce method and remove issue

mrzarquon:
The issue mentioned originally doesn't immediately fix or resolve this problem, it is more for trust to the internal cluster registry if it is used along with self signed certificates

The bruteforce method will work as long as the users don't have any other custom registry settings they want to include

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2534"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2534"></a>


lucasvaltl (approved):
LGTM!

PR #2388, opened 2022-07-07
Add Proof-of-Value Reference Architecture

corneliusludmann:
# Description

This pull request adds the proof-of-value reference architecture (currently GCP instructions only because the AWS instructions for single-cluster are not ready yet).

# Related Issue
- https://github.com/gitpod-io/gitpod/issues/10622

# To Do

- [ ] Add an overview graphic where all components are in the same node including database, object storage, and registry 
      _(could be done in a separate PR as an additional iteration)_

## GCP
- [x] Check which services in the GCP project are needed (e.g. `sqladmin.googleapis.com` is most probably not needed)
- [x] Check if all service account roles are needed
- [x] Test it!

## AWS
- [ ] Once the AWS instructions of the single cluster guide are ready, simplify these similar to the GCP instructions
       _(could be done in a separate PR as an additional iteration)_

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2388"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2388"></a>

PR #2554, opened 2022-08-09
[Single Cluster Ref Arch] Aligns node pool max size across GCP and AWS

lucasvaltl:
## Description
This PR aligns the max size of the node pools for the single cluster reference architecture across AWS and GCP towards: 
- max of 4 nodes for the service node pool
- Max of 50 nodes for the workspaces node pool

## How to test
- Make sure the actual change works: go to netlify > docs> self-hosted>installation-guides>reference-architectures>single cluster reference architecture
- Answer the following questions:
  - Is 4 nodes enough for the services node pool
  - Is 50 nodes enough or too much or just right for the workspace node pool?
  - is this the right approach or should these be different across GCP and AWS?

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2554"></a>


meysholdt (approved):
code changes LGTM

adrienthebo (approved):
LGTM, noting that we'll need to update #2388 accordingly.

gitpod/docs/self-hosted/latest/reference-architecture/_chunks/cluster.md
```diff
@@ -141,7 +141,7 @@ gcloud container node-pools \
     --node-labels="gitpod.io/workload_meta=true,gitpod.io/workload_ide=true,gitpod.io/workload_workspace_services=true" \
     --max-pods-per-node=110 \
     --min-nodes=1 \
-    --max-nodes=50 \
+    --max-nodes=4 \
```
kylos101, line 144: 4 is fine. 

This should be fine for most installs. However, as they increase in size, or meet or exceed similar volume as the saas, they may want to consider changing the node type to `n2d-standard-8`. Especially if they decide to host other apps as services.

For context, with the saass, we only have 3 services nodes (one per zone), but they're bigger (`n2d-standard-8`). They are bigger because we host additional services on the services nodes (IPFS, jaeger, prometheus, autoscaler).

gitpod/docs/self-hosted/latest/reference-architecture/_chunks/cluster.md
```diff
@@ -349,7 +349,7 @@ managedNodeGroups:
     instanceTypes: ["m6i.2xlarge"]
     desiredCapacity: 2
     minSize: 1
-    maxSize: 10
+    maxSize: 50
```
kylos101, line 352: This is a good start, customers may want to tune this, which is fine. For context, I remember when the saas was 30, and 40, and 50. We're increasing to 60 this week for the EU region.

It's more important to ask customers about their workload, to help guide them towards determining a good density (# of regular workspaces per node). If there's no section to socialize that as a thought exercise in the ref arch or guides, I would recommend we plan to do so asap. Going through that exercise early will help everyone better understand the workload so the system can be tuned accordingly.

Once we know about the desired workloads, we can set the memory requests, and that will determine the density given the available memory on the node. A similar need that should be discussed in parallel is the desired storage for `/workspace`. For now, I'd recommend mirroring the saas, with the understanding that some customers may want to change it.

kylos101 (approved):
@lucasvaltl :wave: hey there, these changes look fine. I'll start a separate PR to make some additional improvements. Thanks for putting this together.

PR #2561, opened 2022-08-10
[self hosted] Add sub headings for reference architecture network section

adrienthebo:
## Description

The Network section of the reference architecture has a number of
important sub-sections and expandable setup steps, but the current page
layout doesn't clearly separate or identify what each section is
responsible. This commit adds sub-headings to make section navigation a bit faster.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2561"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2561"></a>


gitpod/docs/self-hosted/latest/reference-architecture/single-cluster-ref-arch.md
```diff
@@ -47,7 +47,7 @@ You can also create a Kubernetes cluster on your bare metal servers for smaller
 
 <Cluster />
 
-## Ingress
+## Network
```
lucasvaltl, line 50: I think this makes sense, but this might break external links that link directly to this section. Likely worth adding a redirect to direct anything that points to `.../single-cluster-ref-arch#ingress` to `.../single-cluster-ref-arch#networking`. This can be done in the _redirects file in the top level of the website repo.

lucasvaltl (approved):
LGTM! Just had a minor comment on redirects :)

gitpod/docs/self-hosted/latest/reference-architecture/single-cluster-ref-arch.md
lucasvaltl: Also, is `Networking` a better term to use instead of network here? I'll leave that to you :)

adrienthebo: I prefer `Networking` over `Network`, will swap that out - and will set up redirects for `#ingress` -> `#networking`. Thanks!

PR #2562, opened 2022-08-10
Fix Svelte/Vite HMR config

felladrin:
## Description

There's an issue preventing us from working with this repository out of the box using VS Code Desktop: the website enters into an endless refreshing loop. ([Internal discussion](https://gitpod.slack.com/archives/C020VCB0U5A/p1660145249168809))

This PR allows it to correctly work on VS Code Browser and Desktop IDEs by forcing Vite HMR websocket to connect to through Gitpod Port URL:

| Running without port forwarding (VS Code Browser) | Running with port forwarding (Desktop IDEs) |
| --- | --- |
| image | image |

## How to test

1. Go to [Gitpod.io Preferences](https://gitpod.io/preferences), select VS Code Browser as Editor
2. Start a workspace using this URL: https://gitpod.io/#https://github.com/gitpod-io/website/pull/2562
3. Connect to the workspace using VS Code Desktop
4. Run this command on terminal: `gp preview --external $(gp url 3000)`, confirm to open the URL in the browser, and check if the website is not in a refresh-loop.
5. In another browser tab, visit http://localhost:3000/ and check if the website is not in a refresh-loop.
6. With these two tabs opened, replace "Always" with "Eternally" on `src/lib/components/index/hero.svelte` and confirm if the website is updated in both tabs.
    | | |
    | --- | --- |
    | image | image |

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2562"></a>


adrienthebo:
I've verified this on the desktop and it works great, but I think that fixing it on the desktop might have broken the browser IDE. Doing more testing to verify this.

adrienthebo (approved):
Verified that this does work on vscode-desktop and vscode-browser. As I briefly mentioned in an earlier comment, because `GITPOD_IDE_ALIAS` is set by the user preferences and not by the currently opened IDE then using vscode-browser to edit the website when the default editor is vscode-desktop won't work - but that's okay, because it's an edge case that I think we can avoid.

This is also a huge usability improvement; I'm heavily dependent on vim keybindings which are only available on vscode-desktop and I'm useless when those bindings disappear. Thanks for tackling this on such short notice! 💥 

/cc @lucasvaltl I believe you ran into this issue as well; do you run into the case where your default editor is vscode-desktop but you use vscode-browser for website edits?

svelte.config.js
```diff
@@ -51,10 +51,12 @@ const config = {
           allow: [".."],
         },
         hmr: {
-          clientPort: process.env.GITPOD_WORKSPACE_URL ? 443 : 3000,
-          host: process.env.GITPOD_WORKSPACE_URL
-            ? process.env.GITPOD_WORKSPACE_URL.replace("https://", "3000-")
-            : "localhost",
+          clientPort: process.env.GITPOD_IDE_ALIAS === "code" ? 443 : undefined,
```
loujaybee, line 54: **Thought(off-topic):** Unrelated to the PR, but this is not a documented env variable, nor guaranteed as an API. I'll make a reference in the following issue (https://github.com/gitpod-io/gitpod/issues/10797) but definitely worth us having a strategy around which env variables can be relied upon.

felladrin:
I am marking it as a draft while I test a new thing that can solve the issues mentioned above.

felladrin:
@adrienthebo @loujaybee
I've updated the solution to avoid using the IDE Alias. Could you please check again?
I'll update the PR description soon.

svelte.config.js
felladrin: Code's been updated to don't make use of the env variables at all :)

felladrin:
I've updated the solution to avoid using environment variables.
Description and instructions for testing were also updated.
Could you please check again?

adrienthebo:
Tested with vs code desktop and browser in parallel; navigating the site was extremely fast and page updates only occurred on file changes. This worked perfectly! LGTM 🚢

akosyakov:
@felladrin it is a right solution 👍

PR #2571, opened 2022-08-11
Add EKS cluster setup guide to proof of value reference architecture

adrienthebo:
## Description

This pull request adds EKS cluster setup instructions to the proof of value reference guide.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/11226

## Release Notes

```release-note
[self hosted] Add EKS reference architecture proof of value guide
```

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2571"></a>


gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -130,9 +130,266 @@ kubectl create clusterrolebinding cluster-admin-binding \
 
 </div>
 
+<div slot="aws">
+
+> **Note:** By default, when Gitpod is being installed using the instructions here, EKS will create a classic load balancer that you can point your DNS entries at. If you are unable to use a AWS Classic Load Balancer (e.g. because you use SSL certificates generated by AWS), please follow [the Setting up your EKS cluster with dual ALB + NLB load balancers guide](../advanced/eks-with-alb-and-nlb) _alongside_ this reference architecture guide.
```
adrienthebo, line 135: Since the proof of value is intended as a simple, quick installation guide the ALB/NLB documentation is likely overkill; we can probably yank this.

lucasvaltl: Agree :)

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -130,9 +135,266 @@ kubectl create clusterrolebinding cluster-admin-binding \
...
+  - name: default
+    amiFamily: Ubuntu2004
+    spot: false
+    instanceTypes: ["m6i.xlarge"]
+    desiredCapacity: 1
+    minSize: 1
+    maxSize: 4
+    maxPodsPerNode: 110
+    disableIMDSv1: false
+    volumeSize: 300
```
lucasvaltl, line 235: We've switched to 512 GB on the single cluster one (for workspace nodes with 8 vCPUs)  - see https://github.com/gitpod-io/website/pull/2630. Might be worth going higher here to something like 768 GB: Using 40 per workspace (default from SaaS) + 5 gb ephemeral storage per workspace (default from SaaS) + some storage for images @ assuming 15 workspaces for this node (we do 18 on SaaS) -->  768 GB ish. Important: This also applies to the GCP version of this, where we currently only provision a single 100 GB disk (!) - given that this has not changed in this PR I cannot comment on that as well.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -130,9 +135,266 @@ kubectl create clusterrolebinding cluster-admin-binding \
...
+
+# Nodegroups / Compute settings
+managedNodeGroups:
+  - name: default
+    amiFamily: Ubuntu2004
+    spot: false
+    instanceTypes: ["m6i.xlarge"]
+    desiredCapacity: 1
+    minSize: 1
+    maxSize: 4
```
lucasvaltl, line 232: We are we capping the size to 4 here and to 50 on the GCP side? Don't get me wrong, 4 is much more reasonable - but we should make sure the two are aligned and not giving different recommendations. I think 4 for both is fine for now given the size of the proposed nodes here.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -16,11 +16,16 @@ title: "Proof-of-Value Reference Architecture"
...
+| Limitations:  | - Not intended for long-term, production usage <br/> - Higher risk of data loss due to in-cluster dependencies - if you lose your cluster, you may lose your data <br/> - Data is not carried over when moving to the [Single Cluster Reference Architecture](./single-cluster-ref-arch) for production purposes |
 
 ## Overview
 
-This reference architecture provides you with a setup where all components that are needed to operate Gitpod are deployed to the Kubernetes cluster. In contrast to a productive setup, this includes a database, an object storage, as well as a registry. For simplicity, we create only a single node where all workload will be deployed to.
+This reference architecture provides you with a setup where all components that are needed to operate Gitpod are deployed within the Kubernetes cluster. In contrast to a production-oriented setup, this means that the required database, object storage, well as a registry are deployed for you within the cluster. For simplicity, we create only a single node where all workload will be deployed to.
```
lucasvaltl: Might be worth clarifying that yes, we create a single node, but that we do allow for autoscaling?

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -130,9 +135,266 @@ kubectl create clusterrolebinding cluster-admin-binding \
...
+
+This should result in:
+
+```
+NAME                       READY   STATUS    RESTARTS   AGE
+coredns-5947f47f5f-79vqv   0/1     Pending   0          22m
+coredns-5947f47f5f-97f8m   0/1     Pending   0          22m
+```
+
+### Calico Installation
```
lucasvaltl, line 304: Currently, there is a UX issue with this headlines in the AWS section of the Kubernetes Cluster subsection. These AWS specific headlines show up in the table of contents on the right hand sinde - however if you are viewing the GCP version, clicking on them does not move you to the relevant section. I believe we get around this in the single cluster ref arch because there we use chunks throughout (i.e. also for the Kubernetes Cluster section), and thus the subheadings are not loaded into the right hand side table of contents. Not sure how to get around this outside of making these not specific headlines, but just bold or something similar.

lucasvaltl (approved):
Overall, LGTM. Did not go through the actual guides and set things up, but rather read through them. Couple of suggestions added.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -130,9 +135,271 @@ kubectl create clusterrolebinding cluster-admin-binding \
...
+      attachPolicyARNs:
+        - arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly
+        - arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy
+        - arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy
+        - arn:aws:iam::aws:policy/ElasticLoadBalancingFullAccess
+        - arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
+
+    tags:
+      k8s.io/cluster-autoscaler/enabled: "true"
+      k8s.io/cluster-autoscaler/gitpod: "owned"
```
adrienthebo, line 252: Bad tag name, replaced by `k8s.io/cluster-autoscaler/gitpod-pov`

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -130,9 +135,266 @@ kubectl create clusterrolebinding cluster-admin-binding \
...
+      gitpod.io/workload_workspace_services: "true"
+      gitpod.io/workload_workspace_headless: "true"
+
+    preBootstrapCommands:
+      - echo "export USE_MAX_PODS=false" >> /etc/profile.d/bootstrap.sh
+      - echo "export CONTAINER_RUNTIME=containerd" >> /etc/profile.d/bootstrap.sh
+      - sed -i '/^set -o errexit/a\\nsource /etc/profile.d/bootstrap.sh' /etc/eks/bootstrap.sh
+```
+
+To ensure there are enough IPs and networking policy enforcement is in place, this proof of value reference architecture uses Calico for networking. To enable Calico in an EKS installation it must be done after the control plane has been provisioned and before the nodegroups have been created.
```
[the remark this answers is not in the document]
adrienthebo: Will handle this in a [follow-up issue](https://github.com/gitpod-io/gitpod/issues/12492) going into the ins and outs of Calico and the tigera-operator.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
adrienthebo: Good point - amended to `All cluster services and workspaces are deployed on a single autoscaling node group.`.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
adrienthebo: Good catch on that - super weird behavior. Headings inside of folded sections don't generate links that'll automatically open the folds so they're not that useful; I've replaced the headings with bold sections.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -130,9 +135,310 @@ kubectl create clusterrolebinding cluster-admin-binding \
...
+
+**`gitpod-cluster.yaml`**
+
+```yaml
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+metadata:
+  name: gitpod-pov
+  region: eu-west-1
+  version: "1.22"
```
liam-j-bennett, line 160: Do we have a plan for when we review the Kubernetes versioning defined here? (This would be a concern in any architecture) My main concern here is with regards to CVEs, but I'm guessing this takes the latest minor version, which alleviates this issue somewhat

adrienthebo: We don't yet have a plan for handling and updating Kubernetes versions, and we'll need to add consideration for GKE and AKS as well. This will install the latest patch revision so we will pick up security updates (and GKE will do the same).

I'll extract this topic to a separate issue as we have a number of follow-up improvements needed across the guides.

geropl (approved):
I didn't test the setup, and am lagging AWS/EKS experience, but reviewed the steps, description and coherence. LGTM!

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
adrienthebo: Issue extracted to https://github.com/gitpod-io/gitpod/issues/12715

PR #2559, opened 2022-08-10
Add AWS cost estimates to Single Cluster Reference Architecture

lucasvaltl:
## Description
Adds AWS cost estimates to the Single cluster reference architecture - GCP estimates are still in progress. 

Also added an estimate as to how many workspaces this may allow you to run, based on [this comment](https://gitpod.slack.com/archives/C01KLC56NP7/p1660154736599959?thread_ts=1660151211.174839&cid=C01KLC56NP7) (this ignores throughput and IOPS)

## Related Issue(s)

Fixes #

## How to Review
Please review the linked [cost estimates](https://calculator.aws/#/estimate?id=dba87a5a0ccb84ae1445eb38c77c170ddc7a819a) (linked from netlify link ---> docs>self-hosted>installation guides > single cluster ref arch) and ensure that:
- These estimates cover everything that costs $$$ in AWS that gets created when this reference architecture is created
- The usage values here make sense:
  - LB traffic: 0.1 GB per hour
  - S3 usage for images & workspace backups: 500mb per image or backup, 200 images and 800 backups per month - 5TB of S3 usage

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2559"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2559"></a>

PR #2574, opened 2022-08-12
Ref arch aws cost estimate improvement - switch to on-demand nodes

lucasvaltl:
## Description
- Changes link to the cost calculator. Cost calculator now better reflects the reference architecture as it uses on-demand nodes. Assumes 100% usage for 12 hours per day. Reasoning: Assuming 12 hours of runtime per day for workspace all nodes. Autoscaling is in place. This covers an 8 hour workday, and has a buffer for when nodes do not scale down to 0 - as as of @Today the min size os the node pool is 1, hence one node will always run. 
- Grammar / wording improvements

--> This is only a small change - it only changes the pricing model used, and aligns it with what is _actually_ used in the reference architecture as of today. As such, there should not be much of a need for a review from everyone. This review already happened in the initial [PR](https://github.com/gitpod-io/website/pull/2559) 
## Related Issue(s)

Fixes #

## How to test
open netlify preview, go to docs>self-hosted>installation guides> reference architectures>single cluster reference architecture --> Look at cost estimates

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2574"></a>


adrienthebo (approved):
LGTM!

PR #2579, opened 2022-08-15
Avatar component improvements

filiptronicek:
## Description
Attempts to make more performant and accessible avatars.

- adds lazy loading to avatars so they don't load on pages like the changelog if not in the viewport (saves potentially megabytes of transferred images)
- by default display only 54px avatars, since we don't need bigger ones and it saves bandwidth
- makes the `title`s easier to read

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Documentation

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2579"></a>

### DIFF
--- a/src/functions/avatar.ts
+++ b/src/functions/avatar.ts
@@ -1,9 +1,12 @@
 import type { Handler } from "@netlify/functions";
-import fetch from "node-fetch";
+import fetch, { Response } from "node-fetch";
 
 const handler: Handler = async (event, _) => {
-  const username = event.queryStringParameters.username;
-  const res = await fetch(`https://www.github.com/${username}.png`);
+  const { username, size = 20 } = event.queryStringParameters;
+  const resLocation: Response = await fetch(
+    `https://www.github.com/${username}.png`
+  );
+  const res = await fetch(`${resLocation.headers.get("location")}?s=${size}`);
   const buffer = Buffer.from(await res.arrayBuffer()).toString("base64");
   return {
     statusCode: 200,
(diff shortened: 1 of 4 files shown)


src/functions/avatar.ts
```diff
@@ -1,9 +1,13 @@
...
-import fetch from "node-fetch";
+import fetch, { Response } from "node-fetch";
 
 const handler: Handler = async (event, _) => {
-  const username = event.queryStringParameters.username;
-  const res = await fetch(`https://www.github.com/${username}.png`);
+  const { username, size = "54" } = event.queryStringParameters;
+  const resLocation: Response = await fetch(
+    `https://github.com/${username}.png`,
+    { redirect: "manual" }
```
Cahllagerfeld, line 8: out of curiosity: I never used `redirect: "manual"`, what's the purpose of it? 🤔

Cahllagerfeld (commented):
Besides that, the changes look good to me 👍

src/functions/avatar.ts
filiptronicek: Hey @Cahllagerfeld!

this line is to prevent the redirection to the actual avatar URL inside of GitHub, because at that point, we don't need the avatar yet. 

We first query (e.g.) `https://github.com/Cahllagerfeld.png`. This just responds with an empty response and a 301 HTTP status. This also includes a header, which we need (`Location`), which is the URL of the avatar itself, which we request in the next request with a custom size (we cannot request the `/username.png` endpoint with a custom size).

PR #2596, opened 2022-08-19
add contents to the compatibility matrix

nisarhassan12:
This is the Pr which is going to be used to fill in the contents to the compatibility matrix added in https://github.com/gitpod-io/website/pull/2595 the base branch for this Pr is that PR it should be first merged into its base PR branch once ready.

There are some placeholder entries in `src/lib/contents/docs/compatibility-matrix.ts` which this Pr is supposed to update with actual data from the various different teams. 

The URL for the product compatibility matrix page is `/docs/product-compatibility-matrix`

-----
### edits after @lucasvaltl added content to this:

@nisarhassan12 I took the liberty of making some changes outside of just adding content. I"m not an engineer, so I hope I did not make things harder for you - if yes, I apologise! :) 

**Changes Made**
- Flipped the icons for the different types of “support”. I did this in a hacky way (introducing new versions to not overwrite the icons used in the feature table on the pricing pages). If we want to change this, I’m happy to do the leg work of implementing a change that is not as hacky.
- Added another optional tool tip next to the status - helps describe in more detail what exactly i supported and what is not, and helps list any limitations. I feel that it makes more sense to have a tooltip next to the status, rather than next to the support policy. It is about the “degree of support”. Left the tooltip next to the support policy as an option because I do feel it might become useful.
- Changed relevance type from `Admin` to `Self-Hosted Admin` - requires a bit of padding improvement - could only make the drop down so-wide.
- Introduced a both setting for relevance. Sometimes there are categories that are relevant for both - these would be filtered out if All is selected. I’m not an engineer, so not sure if I did this right 🙂

**Changes requested / Questions**
 - [x] Can we add more padding between the rows? Also some seem to not be centered (vertically) - e.g. CNI row. Overall should likely increase padding a bit everywhere to give everything more “breathing room”
 - [x] Can we enable links in all columns? . E.g. I want to link out to Github’s docs to show which version of Github Self Hosted is officially supported or link out to other docs sections that explain better what each cetogry is used for (e.g. what the database is used for and why we need it.)

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2596"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2596"></a>


lucasvaltl:
@csweichel and @mrzarquon - it would be great if you could review the contents of this product compatibility matrix here (best done with it rendered in front of you) - open a workspace, open the preview and navigate to docs > reference > compatibility matrix). 
--> The actual contents of this matrix are stored in `src/lib/contents/docs/compatibility-matrix.ts` - please comment on that file, in line, in this PR so we can have targeted discussions. 

I am intentionally only asking the two of you first, in order to keep the amount of reviewers smaller initially and get immediate feedback. I plan to expand reviews to representatives from all teams afterwards. I chose the two of you because:
- @csweichel: you have a great overview of the product, and can thus point out any component or category that is missed or wrong
- @mrzarquon: You are very close to the customer, and I expect that you might find components that are used by customers that are not (or not correctly) represented here.

--> One thing currently still missing is adding codeowners to the `src/lib/contents/docs/compatibility-matrix.ts` file.

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,467 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+        name: "containerd",
         availibility: "supported",
+        limitations: "- Socket access required",
         policy: {
-          text: "the last minor three releases",
+          text: "Last three minor releases",
           description: null,
         },
-        supportedVersions: "1.22, 1.23 and 1.24",
+        supportedVersions: "1.4, 1.5 and 1.6.",
```
csweichel, line 291: I don't think we support containerd 1.4 anymore.
cc @aledbf

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,467 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+    components: [
+      {
+        name: "Ubuntu",
+        availibility: "supported",
+        limitations: null,
+        policy: {
+          text: "All LTS versions and the latest",
+          description: null,
+        },
+        supportedVersions: "18.04, 20.04, 22.04",
```
csweichel, line 202: Not sure we support Ubuntu 18.04 anymore.

cc @gitpod-io/engineering-workspace

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,467 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+          "Cert-Manager is required even if you don't use it for your own certificates because Gitpod uses it internally.",
         policy: {
-          text: "default",
-          description: "",
+          text: "Most recent versions",
+          description:
+            'Any version that implements <a href="https://cert-manager.io/docs/reference/api-docs/#cert-manager.io%2fv1">cert-manager.io/v1</a>',
         },
-        supportedVersions: "default",
+        supportedVersions: "Most recent versions",
```
csweichel, line 325: What's the different between `current version` and `most recent version`?

csweichel (commented):
I love the way this looks - it's very easy to parse.

Regarding content, I left a few comments.
IMHO WebApp, IDE and Workspace should review before making this public. 

The teams need to be aware that there will be support requests for these bits of software!

nisarhassan12:
> - [x] Flipped the icons for the different types of “support”.

@lucasvaltl I have pushed a commit to make the to be consistent see https://github.com/gitpod-io/website/pull/2596/commits/6374b625758fe316d7097a1fa2e457fa1a26ce0b

> - [ ] Can we add more padding between the rows? Also some seem to not be centered (vertically) - e.g. CNI row. 

Overall should likely increase padding a bit everywhere to give everything more “breathing room”

Adding the padding is totally possible I'll do that later on. I have fixed the vertical alignment. 

> - [ ] Can we enable links in all columns?

I just added support for doing so i.e now the authors will be able to insert link tags within the value strings i.e `name: 'Value <a href="/docs">a link</>'` and so on.

src/lib/contents/docs/compatibility-matrix.ts
aledbf: > I don't think we support containerd 1.4 anymore.

We don't, and it's EOL since [March 3, 2022](https://github.com/containerd/containerd/blob/main/RELEASES.md#support-horizon)

Also, we are not using 1.5 in core-dev or SaaS either.

src/lib/contents/docs/compatibility-matrix.ts
aledbf: > Not sure we support Ubuntu 18.04 anymore.

No.

lucasvaltl: Which versions do we support currently then? :)

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Again, which versions do we support then? :)

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: You are right, this is a bit confusing. I had this different because this is not about the current version, but all versions that implement the V1 API spec. Changed this in the most recent commit.

nisarhassan12:
@lucasvaltl I have slightly increased the padding for the rows in the y direction from  `12px` on the top and bottom to `16px` on either side the following is how it looks:
| Before | After |
| - | - |
| image | image |

svenefftinge:
It should go under self-hosted/reference not mixed together with developer-targeted docs.

lucasvaltl:
Hey @svenefftinge, we discussed this via [this](https://www.notion.so/gitpod/7499887c0d9c40a19c61c49868ef60d3) RFC. Please take a look at the arguments there, especially those made by @csweichel at the top towards having a single page for this. As always, happy to reevaluate here if we uncover new arguments for one or the other! :)

svenefftinge:
Ups, I missed the RFC. Still I believe the most important function of documentation is to help the reader. Readers who are interested in installing and operating Gitpod are clearly a different persona than readers who want to run their projects with Gitpod. We already (rightly so IMHO) have two separate sections so we can talk the right audience and assume context. 
Reading through the arguments for a single page, I'd like to propose to link the single page twice. Once in `reference` and once in `self-hosted/reference`. The former link should be parameterized to only select `user` elements by default, the latter should have `all` because the installer person is potentially also a user and/or at least feels responsible for the user's experience.

csweichel:
> Hey @svenefftinge, we discussed this via [this](https://www.notion.so/gitpod/7499887c0d9c40a19c61c49868ef60d3) RFC. Please take a look at the arguments there, especially those made by @csweichel at the top towards having a single page for this. As always, happy to reevaluate here if we uncover new arguments for one or the other! :)

My main concerns with having them on separate pages were:
- some entries can't be clearly separated (e.g. browser compatibility or IDEs)
- if we ran this from two separate sources we'd have a high likelihood of the two lists divering.

I like @svenefftinge's proposal to link to the page with a parameter from separate contexts.
Depending on the effort, we could even make them separate pages fed from the same source down the road.

metcalfc (commented):
I don't have a ton of feedback to add. I think we need to add AWS ACM to the certs component. Because yes you do have to have cert manager but you can do your certs in ACM.

Visually, I had to expand this window fairly wide on a wide format monitor to not have to horizontally scroll the table. If its possible to fit this without scrolling that would be better.

lucasvaltl:
> I'd like to propose to link the single page twice. Once in reference and once in self-hosted/reference. The former link should be parameterized to only select user elements by default, the latter should have all because the installer person is potentially also a user and/or at least feels responsible for the user's experience.

I like this, gives us the best of both worlds. @nisarhassan12 can we make this happen?

src/lib/contents/docs/compatibility-matrix.ts
kylos101:
```suggestion
        supportedVersions: "20.04, 22.04",
```

src/lib/contents/docs/compatibility-matrix.ts
kylos101:
```suggestion
        supportedVersions: "1.5 and 1.6.",
```

I say this only because we use the [unspecified channel](https://github.com/gitpod-io/gitpod/blob/b09a3eb796230b3accbe590358d53b4ddb57e166/install/infra/modules/gke/main.tf#L87) to provision self-hosted infrastructure, [default to using 1.22](https://github.com/gitpod-io/gitpod/blob/2cd2b3e51fadfe151ab334041b4acbe95f2b19ab/install/infra/single-cluster/gcp/variables.tf#L38), and [both 1.22 channels](https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions) support containerd 1.5.

@lucasvaltl How should we communicate that 1.5 should work, but, it is not covered as well as 1.6?
1. integration tests (for workspace) are failing against preview environments, even with 1.6, [we are working to fix and it should also fix for 1.5](https://github.com/gitpod-io/gitpod/issues/12248).
2. integration test results for self-hosted infrastructure are not being shared with product teams with 1.5. Can you share how these are being shared now?
3. as @aledbf indicated, core-dev and saas use 1.6, and we do not support 1.4

kylos101 (commented):
Some feedback for OS and containerd versions

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Happy to merge that in - one question however: Does this not conflict with the current versioning policy here, i.e. "All LTS versions and the latest"? Afaict, there are LTS versions of Ubuntu 14, 16 and  18.

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Happy to merge this and try to state that 1.5 is not tested. This also clashes with the versioning policy here ("Last three minor release"). Should we change the policy?

lucasvaltl:
FYI: I am inviting every Prod eng team to review the contents of `src/lib/contents/docs/compatibility-matrix.ts`  while we improve the UX of this. The content itself should stay the same :)

nisarhassan12:
> @nisarhassan12 can we make this happen?

@lucasvaltl for sure. I'll look into it later this week.

nisarhassan12:
> I'd like to propose to link the single page twice. Once in `reference` and once in `self-hosted/reference`. The former link should be parameterized to only select `user` elements by default, the latter should have `all` because the installer person is potentially also a user and/or at least feels responsible for the user's experience.

@lucasvaltl I have implemented part of this in https://github.com/gitpod-io/website/pull/2596/commits/aaaa3cc58ce83d7b251c8aa27aec1773214f662f i.e now if we add `?user` at the end of the URL it will have the user elements selected by default. Now you'll have to add the 2 links one with `?user` at the end.

lucasvaltl:
Thanks a lot @nisarhassan12 - very cool. I just took the liberty of also adding a deep ling for URLs ending in `?admin` to deep link to the table filtered for Self-Hosted Admins. I am not sure yet if this should be the default or `All` Should be the default of the Self-Hosted "version" of the compatibility matrix.

akosyakov:
Is there a rendered version to read? cc @lucasvaltl

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
     ],
   },
   {
-    name: "Container Network Interface (CNI)",
+    name: "IDE", // RESPONSIBLE TEAM: IDE
     relevance: "users",
     components: [
       {
-        name: "Kubernetes",
+        name: "VS Code Web",
```
akosyakov, line 45: Please use `VS Code Browser` in user facing docs

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
-          text: "default",
+          text: "current version",
           description: null,
         },
-        supportedVersions: "default",
+        supportedVersions: "current version",
       },
       {
-        name: "Elastic Kubernetes Service (EKS)",
+        name: "IntelliJ IDEA",
```
akosyakov, line 65: I'm not sure that it should be here. Backends are part of SH installation, so there is no really requirements on client side. There can be requirements to JetBrains Gateway depending on a version of Gitpod API or SSH Gateway client is running, it is similar to VS Code Desktop.

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+        availibility: "supported",
+        limitations: null,
+        policy: {
+          text: "current version",
+          description: null,
+        },
+        supportedVersions: "current version",
+      },
+      {
+        name: "CLI access (e.g. vim)",
```
akosyakov, line 95: I probably would rather talk about `CLI (SSH)`. Maybe we have requirements to OpenSSH clients running on user machines or keys? @iQQBot Are there any?

src/lib/contents/docs/compatibility-matrix.ts
felladrin: No requirements that I could think of.

For the record, I'm leaving the prerequisites (both for JetBrains Client (Gateway) and JetBrains Backend IDE) links here:

- https://www.jetbrains.com/help/idea/remote-development-a.html#remote_prerequisites
  > Since JetBrains Client is a thin client lightweight application, the requirements for disk space and computer resources are lower than for a host.

- https://www.jetbrains.com/help/idea/installation-guide.html#requirements

jeanp413: No requirements other than having a ssh client installed for the remote-ssh extension to work

akosyakov: @felladrin What about users reporting from time to time that they are broken because of some incompatible Gateway to our backends? Should we put to use latest stable GW to avoid it?

felladrin: > Should we put to use latest stable GW to avoid it?

Ah, well remembered.
As there are two ways of downloading the Gateway (Via Website and via JetBrains Toolbox), I agree we should mention it. But instead of the word "stable", I suggest we use "latest" (meaning the latest available on Website or Toolbox).
I'm saying this because Toolbox allows us to choose between Stable or EAP. But from the website, we can only download EAP.

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Ping on the above in case it got missed :)

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: fixed :)

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Sorry I am not following the conclusion here - what exactly do we want to change here? Alternatively, feel free to make any changes yourself :)

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Changed to CLI (SSH) as suggested. Pinging @iQQBot - what requirements exist here? I can add then in here :)

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: @kylos101  ping on this - should I change the versioning policy to the last two releases? Or fix it at 1.5 and 1.6?

gitpod/docs/references/product-compatibility-matrix.md
```diff
@@ -0,0 +1,19 @@
...
+</script>
+
+<script lang="ts">
+	import CompatibilityMatrix from '$lib/components/docs/compatibility-matrix.svelte';
+
+</script>
+
+# Gitpod Compatibility Matrix
+
+This page shows you what components and services Gitpod is and is not compatible with. It is aimed at both individual developers that use Gitpod as well as admins that install and maintain a [Self-Hosted](../self-hosted/latest) version of Gitpod. You can filter the table to show only what is relevant to you.
```
corneliusludmann, line 17: If we use “self-hosted” as an adjective, I wouldn't capitalize it. Alternativ would something like this:

```suggestion
This page shows you what components and services Gitpod is and is not compatible with. It is aimed at both individual developers that use Gitpod as well as admins that install and maintain a [Gitpod Self-Hosted](../self-hosted/latest) instance. You can filter the table to show only what is relevant to you.
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
   },
   {
-    name: "Image Registry",
-    relevance: "users",
+    name: "Source Control Management System (SCM)", // RESPONSIBLE TEAM: WEBAPP
+    relevance: "everyone",
     components: [
       {
-        name: "Kubernetes",
+        name: "Github.com",
```
corneliusludmann, line 111:
```suggestion
        name: "GitHub.com",
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+          text: "current version",
           description: null,
         },
-        supportedVersions: "1.22, 1.23 and 1.24",
+        supportedVersions: "current version",
       },
       {
-        name: "Elastic Kubernetes Service (EKS)",
-        availibility: "not-supported",
+        name: "Github Enterprise Server",
```
corneliusludmann, line 121:
```suggestion
        name: "GitHub Enterprise Server",
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+        limitations: null,
+        policy: {
+          text: "all officially supported versions",
+          description: null,
+        },
+        supportedVersions:
+          ' <a href="https://docs.github.com/en/enterprise-server@3.6/admin/all-releases#currently-supported-releases" target=”_blank">see currently supported versions</a>',
+      },
+      {
+        name: "Gitlab.com",
```
corneliusludmann, line 132:
```suggestion
        name: "GitLab.com",
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+        availibility: "supported",
+        limitations: null,
+        policy: {
+          text: "current version",
+          description: null,
+        },
+        supportedVersions: "current version",
+      },
+      {
+        name: "Gitlab Self-Managed",
```
corneliusludmann, line 142:
```suggestion
        name: "GitLab Self-Managed",
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+        limitations: null,
+        policy: {
+          text: "current version",
+          description: null,
+        },
+        supportedVersions: "current version",
       },
       {
-        name: "Elastic Kubernetes Service (EKS)",
+        name: "MiniIO ",
```
corneliusludmann, line 365:
```suggestion
        name: "MinIO ",
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+        supportedVersions: "",
+      },
+    ],
+  },
+  {
+    name: "Image Registry", // RESPONSIBLE TEAM: WORKSPACE
+    relevance: "self-hosted admins",
+    components: [
+      {
+        name: "MiniIO backed by AWS S3",
```
corneliusludmann, line 381:
```suggestion
        name: "MinIO backed by AWS S3",
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
+          text: "",
+          description: null,
+        },
+        supportedVersions: "",
+      },
+      {
+        name: "Amazon Elastic Container Registry (ECR)",
         availibility: "not-supported",
+        limitations:
+          'Amazon Elastic Container Registry (ECR) does not implement the <a href="https://docs.docker.com/registry/spec/api/">Docker Registry HTTP API V2</a> spec fully. The spec expects that, if an image is pushed to a repository that doesn’t exist, it creates the repository before uploading the image. Amazon ECR does not do this - if the repository doesn’t exist, it will error on push. We advise to use MiniIO backed by S3 instead.',
```
corneliusludmann, line 394:
```suggestion
          'Amazon Elastic Container Registry (ECR) does not implement the <a href="https://docs.docker.com/registry/spec/api/">Docker Registry HTTP API V2</a> spec fully. The spec expects that, if an image is pushed to a repository that doesn’t exist, it creates the repository before uploading the image. Amazon ECR does not do this—if the repository doesn’t exist, it will error on push. We advise to use MinIO backed by S3 instead.',
```

src/lib/contents/docs/compatibility-matrix.ts
```diff
@@ -2,199 +2,468 @@ import type { Matrix } from "$lib/types/matrix.type";
...
         },
-        supportedVersions: "default",
+        supportedVersions: "",
       },
       {
-        name: "Elastic Kubernetes Service (EKS)",
+        name: "Other Docker Registry API compliant registries",
+        availibility: "untested",
+        limitations:
+          'Any image registry that implements the <a href="https://docs.docker.com/registry/spec/api/">Docker Registry HTTP API V2</a> spec should work - however we cannot guarantee that it will. ',
```
corneliusludmann, line 445:
```suggestion
          'Any image registry that implements the <a href="https://docs.docker.com/registry/spec/api/">Docker Registry HTTP API V2</a> spec should work—however we cannot guarantee that it will. ',
```

corneliusludmann (approved):
Content-wise it looks good to me. Proposed a couple of typo fixes.

src/lib/contents/docs/compatibility-matrix.ts
akosyakov: For JB: latest release of JB GW ideally installed via toolbox that it will auto update
For VS Code Desktop: it should be the same as SSH, ie open ssh client should be installed on a user machine, ideally user should also configure SSH keys for best experience

cc @loujaybee

meysholdt (approved):
I went through the code changes and LGTM!

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Resolving this as this topic was covered via https://github.com/gitpod-io/website/pull/2596#discussion_r959591032

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Implemented a change here that changes both the policy and the currently supported versions to just `20.04 and 22.04` - note that when we do make changes to the version we support here, we will have to update both. This also does not give customers much predictability - but I am fine merging this like this.

src/lib/contents/docs/compatibility-matrix.ts
lucasvaltl: Implemented a change here that changes both the policy and the currently supported versions to just `1.5 and 1.6`. Also added a note making it clear that we prefer 1.6 which is what we use internally. Same as below:  note that when we do make changes to the version we support here, we will have to update both. This also does not give customers much predictability - but I am fine merging this like this.

src/lib/contents/docs/compatibility-matrix.ts
akosyakov: it looks good to me 👍

lucasvaltl:
👋 Merging this now as [communicated](https://gitpod.slack.com/archives/C02J145BU9L/p1662027940555309?thread_ts=1661257022.477759&cid=C02J145BU9L) on slack - everyone should at least have had a look at this :) If there are any more concerns once merged, we can raise another PR :)

PR #2605, opened 2022-08-22
add contact option to the docs widget.

### LINKED ISSUE #2572: [docs] Add contact option to docs feedback widget
**Context**
---
Currently, we have no option to start a dialog with users after they submitted their docs feedback. Some users take a lot of effort to report an issue and we should make sure they feel heard, that we value their feedback and that we keep them updated about changes. 

For that, we want to add a voluntary option to the docs widget to provide the email address and for legal concerns a checkbox that they agree to be contacted.

The decision for this improvement was made [here](https://www.notion.so/gitpod/Docs-Sync-0a2989dbbc874a2e910ce313b63c2a1c#626b674adeac4fdbbc7170685fcaefc6)

**Possible issues we could face**
---
Users submitting their email addresses are expecting to be contacted by us. We should watch out that we don't create false expectations. Based on historic data, around 10% of the feedback includes additional context (Q2: 237 total submissions, ca. 10% would include their email address -> 1-2 support requests per week). Agreed solution: let's try this out and reconsider if we can't handle the support requests.  @Siddhant-K-code volunteered to be the DRI to keep the overhead 🙏 ✨
Let's consider a special labeling for submissions that include an email address so they don't get overlooked in the Slack channel #feedback

**Implementation steps**
---

1. create mockup -> @chrifro 
2. technical implementation (update feedback widget, update Slack integration [I suggest a new channel just for docs-feedback]) -> 
3. DRI to handle support requests -> @Siddhant-K-code 

Would be curious to hear your thoughts on this as well once you are back from vacation @Cahllagerfeld 🙌

nisarhassan12:
Fixes #2572


src/lib/components/docs/feedback-widget.svelte
```diff
@@ -104,25 +107,32 @@
...
-                  type="submit"><span>Send</span></Button
-                >
-              </span>
-            </div>
+            <Input
+              bind:value={email}
+              id="email"
+              name="email"
+              width="100%"
+              placeholder="Email(optional)"
```
Cahllagerfeld, line 117:
```suggestion
              placeholder="Email (optional)"
```

src/lib/components/docs/feedback-widget.svelte
```diff
@@ -104,25 +107,32 @@
...
+              class="text-xs"
+            />
+          </div>
+          <div class="mt-micro">
+            <p class="text-xs mb-x-small">
+              I consent to having this website store my submitted information so
+              that the support team can respond to my inquiry. For more
+              information, see our
+              <a class="link" href="/privacy"> Privacy Policy. </a>
```
Siddhant-K-code, line 126: Since this is not our sales form. So, It might be good to use _`message`_ instead of _`inquiry`_

```suggestion
              I consent to having this website store my submitted information so
              that the support team can respond to my message. For more
              information, see our
              <a class="link" href="/privacy"> Privacy Policy. </a>
```

PR #2624, opened 2022-08-24
Update Single Cluster Ref Arch AWS Workspace Node Storage to max throughput and IOPS

lucasvaltl:
### Changes
1. Change to the max IOPS and throughput: This PR changes the recommended IOPS and throughput configuration for the storage (`gp3`) used for workspace nodes in the AWS version of the single cluster reference architecture to be the maximum possible (see [AWS docs](https://aws.amazon.com/ebs/general-purpose/)) for each. 
2. Increase to 360 GB storage size

## Reasoning
- We are currently expecting users to be able top run 9 Workspaces per node - see (internal) [assumptions](https://www.notion.so/gitpod/b34589d1df2349a6b6a7fe50a06e333f#9c7c3d7233b4429eb4c933722a3728b6)
- Given what we've seen in SaaS, and assuming a similar usage profile to the average SaaS user, the current IOPS and throughput are too low and we are increasing them to be as equivalent as possible to what we are running in SaaS (see related [internal discussion](https://gitpod.slack.com/archives/C02F19UUW6S/p1660144921528479?thread_ts=1660126413.044819&cid=C02F19UUW6S)). 
- This is only applied to workspace nodes because it is assumed that service nodes do not require this amount of throughput or IOPS. 
- Storage size is increased to 360 GB because: With 9 expected workspaces per node, and assuming 40 GB per workspace (30 gb for files, 10 gb for images) we get 360 GB. 

## Important
- This would increase the [currently projected cost](https://calculator.aws/#/estimate?id=d591c32d9457773170ca2e0f68d728fad0e0df19) by 299$ per month (see new [estimate](https://calculator.aws/#/estimate?id=4605b529918bbcde90694ee445e1a5dbd68a6bd9)) (the link to the pricing calculator in the docs page has been updated). 
- If approved, we will need to update the corresponding terraform module @nandajavarma 

## Related Issue(s)

Fixes #

## How to test
- Think through this change and evaluate whether it makes sense
- Ensure how this is done is sufficient - or do we need to change this anywhere else?

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2624"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2624"></a>

PR #2625, opened 2022-08-24
change disk size to 360 workspace nodes - GCP Single cluster Reference Architecture

lucasvaltl:
## Description

**Change:**
- Updates GCP Single Cluster Reference Architecture workspace node disk size to 360 GB (was 100)

**Reasoning:**
- We are currently expecting users to be able top run 9 Workspaces per node - see (internal) [assumptions](https://www.notion.so/gitpod/b34589d1df2349a6b6a7fe50a06e333f#9c7c3d7233b4429eb4c933722a3728b6)
- Storage size is increased to 360 GB because: With 9 expected workspaces per node, and assuming 40 GB per workspace (30 gb for files, 10 gb for images) we get 360 GB.

--> if this gets approved and merged we will need to update the terraform module @nandajavarma :)

## Related Issue(s)
- https://github.com/gitpod-io/website/pull/2624

## How to test
- Think about this change - does it make sense?

## Release Notes

```release-note
NONE
```

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2625"></a>


gitpod/docs/self-hosted/latest/reference-architecture/_chunks/cluster.md
```diff
@@ -129,7 +129,7 @@ gcloud container node-pools \
     create "workload-services" \
     --cluster="${CLUSTER_NAME}" \
     --disk-type="pd-ssd" \
```
adrienthebo, line 131: This is split off from the original pull request, but we're using persistent SSDs for attached storage; this comes out to $0.340 per GB/month; local SSDs cost $0.80 per GB/month. We should see if this persistence is necessary or if our backup/snapshotting is sufficiently resilient to make this unnecessary.

adrienthebo (approved):
LGTM, following up on [this thread](https://github.com/gitpod-io/website/pull/2625/files#r954328020) via slack

gitpod/docs/self-hosted/latest/reference-architecture/_chunks/cluster.md
lucasvaltl: Let's leave the discussion of introducing local SSDs to another PR: https://github.com/gitpod-io/website/pull/2298
--> his PR is more a quick fix to get what we have working better.

PR #1853, opened 2022-04-02
Add export log section in troubleshooting

jeanp413:
## Description

Add export log section in troubleshooting

Related https://github.com/gitpod-io/openvscode-server/pull/327

## Release Notes

```release-note
NONE
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/1853"></a>

PR #2632, opened 2022-08-25
Add `<Keybind>`

filiptronicek:
## Description
This PR introduces a new component, `<Keybind>`. It is exactly what it sounds like - it encapsulates a key combo / keybind / keyboard shortcut for both easier readability as well as more accessible HTML.

### Usage

First, import the component

```ts
import Keybind from "$lib/components/keybind.svelte";
```

Then, encapsulate your keybind within `<Keybind>` like so:
```markdown
To do that, open the editor's Command Palette (<Keybind>CtrlCmd + Shift + P</Keybind>) and search for "Gitpod: Extend Workspace Timeout".
```

this example will produce an output like this:

on macOS

image

on other platforms

image

Note the <kbd>CtrlCmd</kbd> portion of the example. This is one of two special keywords you can use in this component, which will make them shown up as the correct platform key.

| Name | Win / Linux | macOS |
|---|-----|----|
| <kbd>CtrlCmd</kbd> | <kbd>Ctrl</kbd> | <kbd>⌘</kbd> |
| <kbd>AltOption</kbd> | <kbd>Alt</kbd> | <kbd>⌥</kbd> |

### Considerations

- what should be displayed on mobile devices? Ctrl or Cmd? Or should it again depend on the platform there?
- is this not too complicated of an approach for simply adding keybinds to a markdown document? 

### Further reading / resources

- the `<kbd>` element on MDN [[1](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd)]

### Ideas for the future

I would love to see a similar component for UI elements users are expected to click, like menu items, which the same `<kbd>` element is also used for (along with `<samp>`) [[spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-kbd-element)]. 

## Related Issue(s)
Originally inspired by https://github.com/gitpod-io/website/pull/1853#discussion_r841093631.

## How to test
Look at a page like Troubleshooting and see whether the keybind component renders correctly.

## Release Notes

```release-note
NONE
```

## Documentation
Maybe update the [internal blogging docs](https://www.notion.so/gitpod/Blogging-34c2e078569b4174b42784854f42549f). 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2632"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2632"></a>


Cahllagerfeld (approved):
Looks good to me 👍

src/lib/components/keybind.svelte
```diff
@@ -0,0 +1,42 @@
...
+    if (slotData) {
+      keys = slotData.innerText
+        .trim()
+        .split("+")
+        .map((key) => key.trim())
+        .map((key) => {
+          let isMac = false;
+          if (!navigator.userAgent.toLowerCase().match(/mobile/i)) {
+            isMac = /(Mac)/i.test(navigator.userAgent);
+          }
```
jeanp413, line 14: vscode uses `Macintosh` => `_userAgent.indexOf('Macintosh') >= 0;` is it better?
What about mobile, I don't have an iphone so don't know if it should apply to them too? for that vscode uses `_userAgent.indexOf('Macintosh') >= 0 || _userAgent.indexOf('iPad') >= 0 || _userAgent.indexOf('iPhone') >= 0) && !!navigator.maxTouchPoints && navigator.maxTouchPoints > 0;`

filiptronicek: Thanks for the feedback here, @jeanp413 🙏. I decided to match `Macintosh` as you recommended. For now I'd leave iPhones to display <kbd>Ctrl</kbd>, because 1. keyboard shortcuts aren't that important on mobile and 2. statistics:
> While almost all ‌iPhone‌ users have a computer, only a 41 percent share of these users have a Mac, with the majority having a Windows or Google Chrome device instead. [[1](https://www.macrumors.com/2021/08/25/iphone-serves-as-gateway-to-ipad-and-apple-watch/)]

filiptronicek: For iPads I think it is a different story though. Many folks decide to use a keyboard with that thing and there you can find the standard Command and Option keys. Included them too.

jeanp413 (approved):

PR #2408, opened 2022-07-12
Adapt accessiblity updates from IDE toggle into Cloud platform toggle

filiptronicek:
Companion to #2302. Maybe should wait until #2324 lands, @lucasvaltl? 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2408"></a>

PR #2635, opened 2022-08-25
Improve a11y on navigation

filiptronicek:
## Description
Adds various HTML semantics + ARIA roles so that screen readers can better comprehend our navigation. More of these PRs are to follow, because to be fully up-to-spec will take a lot more effort, although shipping this skateboard will help quite a bit 🤞.

## Related Issue(s)

This is a part of a11y efforts like https://github.com/gitpod-io/website/pull/2632, https://github.com/gitpod-io/website/pull/2408 or https://github.com/gitpod-io/website/pull/2302.

## How to test
Go over the navigation with an assistive technology like [VoiceOver](https://support.apple.com/en-asia/guide/voiceover/welcome/mac) and see how usable it is if you couldn't use a mouse.

## Further reading

- A good guide for accessible menus: https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/

## Release Notes

```release-note
NONE
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2635"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2635"></a>


src/lib/components/main-nav/nav-item.svelte
```diff
@@ -21,18 +21,24 @@
...
+    class:active={isActivePage && !isExternal}
+    class:flex={highlight}
+    {href}
+    on:click
+    sveltekit:prefetch={isPrefecthable}
+    class="text-important sm:text-body text-p-large hover:text-important focus:text-important active:text-important"
+  >
+    {label}
+    {#if highlight}
+      <img src="/indicator.svg" alt="" class="h-1.5 w-1.5" />
```
filiptronicek, line 41: since this image is only decorative, it shouldn't have an `alt` description [[1](https://www.w3.org/WAI/tutorials/images/decorative/)]. 

The fact that this is purely decorative is more of me just guessing, but I do not see how this little dot could have much meaning to users of assistive technologies.

src/lib/components/main-nav/index.svelte
```diff
@@ -79,24 +78,21 @@
...
-        $menuState = !menuState;
-        showHideOverflowY(false);
-        goto("/");
-      }}
+      href="/"
     >
       <Logo class="h-8 w-28 lgx:h-10 lgx:w-32" />
-    </button>
-    <div
+    </a>
```
filiptronicek, line 94: Maybe there was a good reason of using a `<button>`, but an `<a>` should be used here since it is just a link.

src/lib/components/main-nav/index.svelte
```diff
@@ -79,24 +78,21 @@
...
+    <ul
```
nisarhassan12, line 95: Many Thanks, @filiptronicek for taking the initiative to work on this. Can you explain what semantic does it adds and what is the drawback of having a `div` here? 

Also on line number 104 we have the `<Dropdown />` component if we keep this change then we'll have to wrap that as well in a `li`.

src/lib/components/main-nav/index.svelte
nisarhassan12: The current behaviour on the live site is that when you're on the homepage and the mobile menu is opened and clicking the logo closes that but I think it is fine to leave that out. Good call with converting it to a link 👍🏿

src/lib/components/main-nav/nav-item.svelte
nisarhassan12: Thanks for bringing this up. I would say it is not decorative and should have an alt to indicate that the page that it links to has some new content or important updates. This is how I perceive it by looking at it visually as a user. Feel free to suggest if you know what can be good here for the alt text.

src/lib/components/main-nav/index.svelte
```diff
@@ -79,24 +78,21 @@
   class="fixed z-40 mx-auto w-full border-b border-solid border-transparent border-t-0"
   class:scrolled-out={scroll > 0}
   class:bg-open-state={$menuState}
+  aria-label="Main"
```
nisarhassan12, line 81: Can you explain what purpose this serves here? Thanks

src/lib/components/main-nav/index.svelte
filiptronicek: @nisarhassan12 I chose to use a list here, since a `<div>` is by itself really inaccessible. It can be so many different things so using an unordered list here makes it clear that the subset of actionable items in the navigation which is the links is a list of items.

Thanks for bringing up the `<Dropdown />` menuitem. I will wrap it like you suggested, but this one will need a follow-up PR on its own, it will need some more work.

src/lib/components/main-nav/index.svelte
filiptronicek: This announces the navigation as "Main navigation", so the user is aware of the fact this is the menu that interacts with the top-level website links. The sidemenu we have on the left is going to have a different label to make its purpose clear.

nisarhassan12: Thanks @filiptronicek for the explanation.

src/lib/components/main-nav/nav-item.svelte
filiptronicek: Good point, I see what you mean. I still believe the image by itself does not add much meaning, I think we should instead add an `aria-label` to the link itself like "Highlighted page: Customers", so that it is easier to comprehend.

src/lib/components/main-nav/mobile-menu/index.svelte
```diff
@@ -109,5 +109,5 @@
         <LoginButton />
       {/if}
     </div>
```
nisarhassan12, line 111: @filiptronicek we can't use a div here on line number 101 because a `ul` can't have any other direct child other than the `li` element. 

Even if we were to convert this into a list we will get into a bit of a tangle i.e the first two children of this div are links while the third one is now a li element containing a link. 

This all sort of breaks the semantics in a way i.e we can't have a nested ul here because the link buttons are links with the same importance as the other links(NavItems) i.e they are not a sublist and so on.

I'm wondering what can be the solution here. WDYT? Thanks :)

filiptronicek: Good point, thanks! Reverting the commit.

filiptronicek: We can revisit this in the future to address accessibility on mobile as well.

nisarhassan12:
Many Thanks, @filiptronicek for the PR. LGTM 👍🏿 I'll now go ahead and merge it 🚀

PR #2636, opened 2022-08-25
[ref arch/eks] Use Cloudflare, Google DNS servers for dns01 self check

adrienthebo:
## Description

As described in the dns01 challenge provider documentation,
cert-manager will query for challenge TXT records before updating any records to
avoid work duplication. However some recursive DNS servers will cache
negative DNS lookups. In this scenario cert-manager may become stuck by
querying for challenge TXT records, getting a negative result,
performing the record update, requerying, and still retrieving those
cached negative lookups.

At present we avoid that risk by using fast recursive DNS servers hosted
by Cloudflare and Google, but only do so on GKE; the EKS guide does not
replicate this step and thus runs the risk of triggering this bad
behavior.

This commit resolves this issue by applying the same self-check
recursive servers on EKS as we use on GKE.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12413

## How to test

On a reference EKS cluster, apply the given helm chart and confirm that the arguments are passed to the cert-manager pod with the following:

```bash
kubectl --namespace cert-manager get pod -l app=cert-manager -ojson | jq '.items[0].spec.containers[0].args'
```

## Release Notes

```release-note
Use fast dns01 self check nameservers
```

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2636"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2636"></a>


lucasvaltl:
Overall this looks good to me - but I'm probably not the best person to review this, as I am not an expert on the technical details here.

mrzarquon (approved):
✅

PR #2637, opened 2022-08-25
[ref-arch/eks] Update ClusterIssuer example to follow best practices

adrienthebo:
## Description

This commit adjusts the EKS ClusterIssuer example to call out potential configuration issues and follow patterns used earlier in the EKS guide, as follows:

- The `$LETSENCRYPT_EMAIL` variable is used in place of a static string to call attention that users must adjust this value for their environment
- The `$DOMAIN_NAME` variable is used in place of a static domain to match the variable usage followed in other sections of the guide and to call attention that this value must be adjusted to the user environment
- The route53 hostedZoneID value has been dropped as the route53 dns01 challenge provider can infer the correct hostedZoneID, and the eksctl cert-manager wellKnownPolicies setting will grant cert-manager the appropriate rights to enumerate the correct hostedZoneID. Allowing the route53 dns01 challenge provider avoids potential injection of the wrong hostedZoneID which can cause cert-manager issues that are tricky to isolate.
- A callout has been added cautioning around the behavior of the route53 region. This issue is particularly nasty because Route53 is one of two AWS APIs that are bound to specific AWS regions and behave very badly when set to any other region.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12399
See also: https://github.com/gitpod-io/website/pull/2636

## How to test

1. Create an EKS cluster following the reference architecture
2. Create a ClusterIssuer that uses an AWS region that doesn't support route53 calls (such as `us-west-1`); Install gitpod and confirm that installation fails.
3. Amend the ClusterIssuer to use the `us-east-1` region, install Gitpod, and verify that installation succeeds.

If this issue can be reproduced prior to the fix and confirmed as fixed with the region change, we should look very closely at adding a preflight check (per https://github.com/gitpod-io/gitpod/issues/12011) because it'll be easy for unsuspecting AWS users to get mired in this problem.

## Release Notes

```release-note
```


lucasvaltl:
LGTM :)

PR #2649, opened 2022-08-26
[ref-arch/eks] Add external-dns helm chart for EKS

adrienthebo:
## Description

At present the GKE reference architecture uses external-dns to manage DNS records for Gitpod services, but the EKS guide doesn't have a similar system. It has been assumed (but not well documented) that end users would manage Gitpod DNS records manually but this has proven more painful than desired, especially when stale records linger after Gitpod is wiped and reinstalled.

This commit adds documentation for creating an external-dns Kubernetes service account and binding Route53 IAM permissions to that service via `eksctl`, and installing the external-dns helm chart that depends on this pre-existing service account.

It should be documented that this step is optional and end users can substitute their own DNS records management. But as the reference architectures define our supported mechanism for installing Gitpod, making this part of the default procedure puts users in a situation where they'll get a functioning cluster more quickly and then can tailor it to their needs rather than defaulting to manual DNS records creation.

## Related Issue(s)

Fixes https://github.com/gitpod-io/gitpod/issues/12395

## How to test

Run through the EKS setup procedure and verify that upon installing Gitpod the necessary A records have been created with the following:

```bash
sudo apt install dnsutils
dig -t A dig gitpod.$DOMAIN_NAME
dig -t A dig *.gitpod.$DOMAIN_NAME
dig -t A dig *.ws.gitpod.$DOMAIN_NAME
```

## Release Notes

```release-note
[self-hosted] Add external-dns service to EKS reference architecture.
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2649"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2649"></a>


gitpod/docs/self-hosted/latest/reference-architecture/_chunks/networking.md
```diff
@@ -151,6 +151,54 @@ Which should return something like:
...
+
+```bash
+HOSTED_ZONE_ID="$(aws route53 list-hosted-zones-by-name \
+    --dns-name "$DOMAIN_NAME." \
+    --query "HostedZones[0].Id" \
+    --output json \
+    --out text)"
+```
+
+Then install external-dns so that DNS records will be automatically created for Gitpod services.
```
lucasvaltl, line 164:
```suggestion
Then install external-dns so that DNS records will be automatically created for Gitpod services. This can be ignored if you are managing DNS records yourself. 
```

lucasvaltl (approved):
LGTM - but I did not test this from a technical perspective :)

gitpod/docs/self-hosted/latest/reference-architecture/_chunks/networking.md
adrienthebo: 💯

PR #2654, opened 2022-08-29
fix: code blocks styling

Siddhant-K-code:
fixes [internal discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1661282709747239)

---

New Look for Code Blocks

image

- Test it Here on [Netlify Preview](https://deploy-preview-2654--gitpod-kumquat.netlify.app/guides/guide-ephemeral-dev-environment-on-gitpod)

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2654"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2654"></a>


chrifro:
Thanks for looking into this @Siddhant-K-code 
Can we keep the styling as it is for desktop view and only apply the new styling for smaller devices? e.g. when the font size changes from 18 to 16px we could apply this change as well

Siddhant-K-code:
> Can we keep the styling as it is for desktop view and only apply the new styling for smaller devices? e.g. when the font size changes from 18 to 16px we could apply this change as well

I think the current style looks weird on Desktop too if we have consecutive lines of code blocks. Some of the parts get overlapped. I think this change would be ideal for both desktop & mobile.

current live site: 
image


After this PR merge (see that little overlap fix b/w `*/dropwizard` & next line `dropwizard`:
image

chrifro:
Thanks for highlighting it. Let's change it globally but let's keep the new style as close as possible to the previous style (same border-radius, same width spacing and only reduce the spacing in height)
Bildschirmfoto 2022-08-31 um 07 59 56
Bildschirmfoto 2022-08-31 um 08 02 09

Siddhant-K-code:
Resolved in [`c881e1c`](https://github.com/gitpod-io/website/pull/2654/commits/c881e1c381ae9d6d5dfcf0bde29c5c61333dd471)

chrifro (approved):

PR #2664, opened 2022-08-30
[ref-arch/eks] Add cluster autoscaler to EKS PoV reference architecture

adrienthebo:
## Description

This commit stubs out autoscaling for the EKS proof of value reference architecture. As there are multiple sources of truth for cluster autoscaling on EKS the flags passed to the autoscaler deserve closer inspection.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2664"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2664"></a>


gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -366,6 +366,32 @@ K8S:.status.nodeInfo.kubeletVersion,\
...
+    --reset-values \
+    --wait \
+    --set cloudProvider=aws \
+    --set awsRegion=$AWS_REGION \
+    --set autoDiscovery.clusterName=$CLUSTER_NAME \
+    --set serviceAccount.create=false \
+    --set serviceAccount.name=cluster-autoscaler \
+    --set securityContext.fsGroup=65534 \
+    --set extraArgs.skip-nodes-with-local-storage=false \
+    --set extraArgs.skip-nodes-with-system-pods=false \
```
adrienthebo, line 388: Care may be warranted that the cluster autoscaler can't evict itself and effectively disable autoscaling. More information will be added as we understand more.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -366,6 +366,32 @@ K8S:.status.nodeInfo.kubeletVersion,\
...
+helm upgrade \
+    --atomic \
+    --cleanup-on-fail \
+    --install \
+    --namespace kube-system \
+    --reset-values \
+    --wait \
+    --set cloudProvider=aws \
+    --set awsRegion=$AWS_REGION \
+    --set autoDiscovery.clusterName=$CLUSTER_NAME \
```
adrienthebo, line 383: This variables aren't defined in the reference architecture; we should expect that people will be copy/pasting from the RA and should have these set accordingly.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -366,6 +366,32 @@ K8S:.status.nodeInfo.kubeletVersion,\
 Instance-ID:.spec.providerID"
 ```
 
+### Configure cluster autoscaling
+
```
adrienthebo, line 370: A note should be added indicating that we're using the eksctl provided service account.

gitpod/docs/self-hosted/latest/reference-architecture/proof-of-value.md
```diff
@@ -366,6 +366,32 @@ K8S:.status.nodeInfo.kubeletVersion,\
...
+    --install \
+    --namespace kube-system \
+    --reset-values \
+    --wait \
+    --set cloudProvider=aws \
+    --set awsRegion=$AWS_REGION \
+    --set autoDiscovery.clusterName=$CLUSTER_NAME \
+    --set serviceAccount.create=false \
+    --set serviceAccount.name=cluster-autoscaler \
+    --set securityContext.fsGroup=65534 \
```
adrienthebo, line 386: We've needed to set the fsGroup for other pods to allow access to either the Kubernetes service account token or the AWS credentials, but this is a bit of cargo culting. More work is warranted to understand if this is necessary.

adrienthebo: This _might_ not be necessary if the cluster-autoscaler pod runs as root (within the pod) and thus avoids filesystem permission errors. If so then we need to document this behavior.

PR #2611, opened 2022-08-23
fix: 404 page render in prod.

Siddhant-K-code:
## Description

This PR aims to fix #1479. So, If any user will redirect to any invalid URL. It will be redirected to 404 Page.

## How to Test
- Open this Invalid URL: https://deploy-preview-2611--gitpod-kumquat.netlify.app/invalid-random-url

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2611"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2611"></a>

PR #2667, opened 2022-08-31
add redirects to fix 404 custom page error

Siddhant-K-code:
addition to #2611 (due to [Internal Discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1661936479110809))
---

## Description

- This Page adds up a custom page component for 404 and a file in routes.
- addition of a redirect.

## Test
Go to -> [Invalid Link](https://deploy-preview-2667--gitpod-kumquat.netlify.app/invalid-link). It will automatically redirect to `/404`

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2667"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2667"></a>


Siddhant-K-code:
```bash
3:26:03 PM: > Using @sveltejs/adapter-netlify
3:26:03 PM: > Redirects are not supported in netlify.toml. Use _redirects instead. For more details consult the readme's troubleshooting section.
```

we can't use redirects in `netlify.toml`

_redirects
```diff
@@ -1,3 +1,4 @@
+/*    /*  404
```
Cahllagerfeld, line 1:
```suggestion
/*    /404  404
```
To make that working we would need to add a custom `404.svelte` in the routes directory to generate that route. This would probably need to be unindexed from SEO
I'm not sure though if that needs to be the very last entry in `_redirects`.

Cahllagerfeld (commented):
I think we should remove that page from the sitemap + and apply `<meta name="robots" content="noindex">` in the header, so it's unindexed from SEO

Cahllagerfeld (approved):
Looks good to me 🚀

PR #2700, opened 2022-09-08
Remove Slack integration

Cahllagerfeld:


package-lock.json
```diff
@@ -798,6 +776,7 @@
     },
     "node_modules/@types/node": {
       "version": "14.14.35",
+      "dev": true,
```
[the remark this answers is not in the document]
Cahllagerfeld: Its autogenerated..

docs/adr/0002-remove-slack-integration.md
```diff
@@ -0,0 +1,20 @@
...
+Context here...
+
+## Decision
+
+Remove the Slack Integration from the Feedback Endpoint
+
+## Consequences
+
+1. We can get rid of the dependency `@slack/webhook`
+1. The Feedback get send to Slack as a zap, that gets triggered by the Google Spreadsheet
```
Siddhant-K-code, line 20:
```suggestion
- We can get rid of the dependency `@slack/webhook`
- The Feedback gets sent to Slack as a zap, that gets triggered by the Google Spreadsheet
```

docs/adr/0002-remove-slack-integration.md
```diff
@@ -0,0 +1,20 @@
...
+
+Date: 2022-09-08
+
+## Status
+
+2022-09-08 accepted
+
+## Context
+
+Shutting down the Slack Integration in favor of a zap caused the feeback form to return `404` responses (as the webhook didn't exist anymore).
```
Siddhant-K-code, line 11:
```suggestion
> [Internal Discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1662639377730739)

Shutting down the Slack Integration in favor of a zap caused the feedback form to return `404` responses (as the webhook didn't exist anymore).
```

Cahllagerfeld: I'm not sure yet, if I'd add an internal discussion to an adr, but in case we don't like it, we can remove it again in the future 🤔

Siddhant-K-code (approved):

PR #2702, opened 2022-09-08
Changelog images rounded corners

Cahllagerfeld:


Siddhant-K-code (approved):

src/lib/components/changelog/wrapper.svelte
```diff
@@ -3,6 +3,8 @@
   export { clazz as class };
 </script>
 
-<div class="content-docs prose prose-h3:mb-6 max-w-none {clazz}">
+<div
+  class="content-docs prose-img:rounded-t-2xl prose prose-h3:mb-6 max-w-none {clazz}"
```
nisarhassan12, line 7: I think we should use a smaller radius value for mobile maybe `rounded-t-md`? because `rounded-t-2xl` is too much:

image

PR #2743, opened 2022-09-15
reduce the number of css chunks.

### LINKED ISSUE #1997: Performance (number of css chunks)
### Page URL

https://www.gitpod.io/

### User feedback

Because of how sveltekit works, there is 16 css files that sum up to 20KB (compressed).
Most of them are chunks from components, that have less than 1KB, which means HTTP request is a very heavy tool for loading them.

You can minimize chunk generation by not using `<style>` tags when tailwindcss classes can do the same job.
Example: https://github.com/gitpod-io/website/blob/main/src/lib/components/index/testimonial.svelte#L8-L30

Report: https://webpagetest.org/result/220508_BiDcCB_5XS/1/details/#waterfall_view_step1
**Notice** - The orange crosses next to CSS files, they delay rendering of the whole page.

Apart from minimizing chunks css that are very small, I would recommend inlining the font.css and style.css files since they are so small.

Those operations should make your site start rendering a little bit quicker.

nisarhassan12:
Fixes #1997 

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2743"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2743"></a>


pavelloz:
🥳

PR #2746, opened 2022-09-16
fix: update free-plan benefit to 100h

depapp:
## Description

I'm on `Free Plan` and got `100 hours` benefit. so I think the docs that related to `50 hours` need updated to `100 hours`.
my changes are update `50h` to `100h`
Screen Shot 2022-09-16 at 08 30 57

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2746"></a>


Siddhant-K-code:
Thanks for raising the PR ✨ 

But, This depends on their signup date.

[source](https://github.com/gitpod-io/gitpod/blob/78a0bf50a30fd08e24c5534b553c8e7fc6c446b5/components/gitpod-protocol/src/plans.ts#L158)

until that date, users get 100 free hours when signing up, and we grandfathered them with that.
All newer users get 50h free hours.

> **Note** that this will likely change with Usage-Based Pricing.

PR #2757, opened 2022-09-20
Fix Layout Endpoint

Cahllagerfeld:
Fixes [Internal Discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1663357972724979)


Cahllagerfeld (commented):
@Siddhant-K-code I think the issue was, that `bannerData` was initialized in an if-block. I moved it out that block

Siddhant-K-code (approved):

PR #2760, opened 2022-09-20
September Changelog

loujaybee:
> **Note** Only make changes to this PR to add ad-hoc changelog content, the automated list will be re-generated. To update the changelog list of items, update the release notes section on your PR directly. To omit, simply add `NONE` to the release notes field. 

Script for re-generating: `node scripts/generate-changelog.js 2022-09-30 2022-09-01 2022-09-30`

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2760"></a>

**Features to highlight:**
- [x] [New ports view for VS Code Browser](https://www.gitpod.io/changelog/new-ports-explorer-changelog-view-and-connection-improvements-for-vs-code#new-ports-explorer-in-vs-code-browser)
- [x] [JetBrains common actions](https://github.com/gitpod-io/gitpod/issues/12884)
- [x] JetBrains workspace class information
- [x] JetBrains reconnecting windows
- [x] Alt text for images added
- [x] Copy images to current repo (not external)
- [x] SSH Gateway default for VS Code Desktop users in Self-Hosted
- [x] @jacobparis update header image

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2760"></a>

PR #2762, opened 2022-09-20
fix: null & undefined errors

Siddhant-K-code:
fixes [internal discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1663671454861139?thread_ts=1663357972.724979&cid=C01KGM9ETU6), also [[1]](https://app.netlify.com/sites/gitpod-kumquat/deploys/63298de41444a40009a55cb9), [[2]](https://app.netlify.com/sites/gitpod-kumquat/deploys/632968f717872600089f4a7a)

Signed-off-by: Siddhant Khare <siddhant@gitpod.io>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2762"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2762"></a>


src/lib/utils/helpers.ts
```diff
@@ -6,7 +6,7 @@ export const isEurope = () => {
 };
 
 export const stringToBeautifiedFragment = (str: string = "") =>
```
Cahllagerfeld, line 8: I think this function already has `str` set to an empty string as default value 🤔

Cahllagerfeld:
Except for the build that failed for the changelog PRs, I didn't see any more failures. The one from #2760 can be fixed by filling out `title` and `excerpt` in the markdown frontmatter. In my workspace, the build was succeeding after that.

Cahllagerfeld:
I think it may be better so search for the actual issue, instead of adding fallback values in multiple places 🤔 Wdyt?

Siddhant-K-code:
we still got the same error [here](https://app.netlify.com/sites/gitpod-kumquat/deploys/632c542e3a21ed0008f27ebb#L1655) in [`this commit`](https://github.com/gitpod-io/website/commit/3db3667d9b16e97a15a51125b3d54d304b033296)

src/lib/components/main-nav/github-stars.svelte
```diff
@@ -6,7 +6,7 @@
   import { getContext } from "svelte";
   import Octocat from "../svgs/octocat.svelte";
 
-  const stars: string = getContext(key).toString();
+  const stars: string = getContext(key).toString() || "9089";
```
Cahllagerfeld, line 9: To not waste too much time on it, I just change the fallback value to an actual number. I think we can schedule a cleaner solution, once we have more time. We should focus on the docs-restructuring for now. What do you think? @Siddhant-K-code

Siddhant-K-code: > We should focus on the docs restructuring for now. 

I am already on it.

> I just change the fallback value to an actual number

what is this `9089`?

> As we know the reason for the flaky builds now

can you pls quickly tell me about your findings?

Cahllagerfeld: > can you pls quickly tell me about your findings?

https://gitpod.slack.com/archives/D03UMJ97KC3/p1663859258811849

> what is this 9089?

It's the current amount of stars the main repo has. We display this number in case the api request for the given reasons in the slack thread fails

Siddhant-K-code: > It's the current amount of stars the main repo has. We display this number in case the API request for the given reasons in the slack thread fails

I think we should make it "Stars"? instead of all that random nums. because it is not constant.

Issue #2721, opened 2022-09-12
[Docs Feedback]: Improve operational instructions for Gitpod update


kylos101:
1. Share `Live update (No downtime, higher risk, low cost)` for Single Cluster Ref Arch is not supported.
2. Share `maintenance window` must start by checking to assert that there are no running workspaces. This is a :skateboard: , for the future, we'll want a means to "enable" maintenance windows via webapp or perhaps using `gpctl` to temporarily disable routing of start workspace requests.

### Page URL

https://www.gitpod.io/docs/self-hosted/latest/updating#live-update-no-downtime-higher-risk-low-cost 
https://www.gitpod.io/docs/self-hosted/latest/updating#maintenance-window-higher-downtime-low-risk-and-medium-cost---recommended

### User feedback

**Live update**
If a user upgrades a cluster while workspaces are live, and the ws-daemon restarts while a workspace is stopping (backing up) they could encounter data loss. This is why this approach is considered high risk, and why a maintenance window is required. In other words, Single Cluster Ref Arch in its current form does not support "No downtime" upgrades.

In other words :point_up: is not an edge case when workspaces are running. The more workspaces running during the upgrade, the greater there is a chance of data loss, and having to do related clean-up of the workspaces.

**Maintenance window**
This page simply needs to be updated to inclube verbiage that indicates the admins must check for running workspaces (imagebuild, prebuild, and regular) before starting the upgrade.

### Definition of done

The documentation is updated to clearly outline that No download upgrades for Single Cluster Ref Arch is not supported. The related PR must be approved by the CX and TS teams.

kylos101:
@jldec hey, are there any plans for webapp to include the capability to enable maintenance windows, and avoid sending start workspace requests for imagebuild/prebuild/regular? If so, can you share roughly when after UBP?

PR #2777, opened 2022-09-23
Prevent workspace starts during Gitpod update for single cluster ref arch

kylos101:
## Description

If we do not recommend this, and there are running workspaces when Gitpod is updated, users could incur data loss.

## Related Issue(s)

Fixes: https://github.com/gitpod-io/website/issues/2721

## How to test

* Stop running workspaces
* Scale-down ws-manager to zero
* Try starting a workspace

## Release Notes

```release-note
Prevent workspace starts during Gitpod update for single cluster ref arch
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2777"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2777"></a>


kylos101:
Will mark ready for review after running a brief test upgrading from 7.1 to 8.0.

gitpod/docs/self-hosted/latest/prevent-workspace-starts.md
```diff
@@ -0,0 +1,27 @@
...
+  export const prerender = true;
+</script>
+
+# Prevent Workspace Starts
+
+It may be necessary for you to prevent users from starting Gitpod workspaces (in a `<namespace>`). Please see the [update guide](./updating) for more information.
+
+> It is required to prevent workspace starts when updating Gitpod and using the [Single Cluster Reference Architecture](../latest/reference-architecture/single-cluster-ref-arch) because it is not highly-available.
+
+This can be done with `kubectl`, after which, users won't be able to start workspaces until the `ws-manager` pod is running again.
```
Furisto, line 17:
```suggestion
This can be done by scaling the ws-manager component down to zero, after which, users won't be able to start workspaces until the `ws-manager` pod is running again.
```

lucasvaltl (changes requested):
This looks good to me - I've not run the command (kubectl scale --replicas=0 deployment/ws-manager -n <namespace>) to test it in practice though. 

One question - how would a user then get back into a "normal" state post update? Does this happen by default, because the update will overwrite the change to the ws-manager replicas set by the command? It might be worth explaining this if this is not a super obvious thing that I am just unaware of (not a K8s expert) :) 

Other than this, happy for this to be merged without another review.

Furisto:
You know this better than me but as far as I understand an update deletes the workspace components and then recreates them. In that case nothing would have to be done because the new deployment will have the correct replica count.

Pothulapati:
>update deletes the workspace components and then recreates them. In that case nothing would have to be done because the new deployment will have the correct replica count.

Yep, A `kots` update does that i.e removing existing install and re-appllying a new version which means the coutn goes back to the default.

> the update will overwrite the change to the ws-manager replicas set by the command? It might be worth explaining this if this is not a super obvious thing that I am just unaware of (not a K8s expert)

A note makes sense! in the lines of "After you do this, please know that an upgrade is a requirement to make the `ws-manger` working again"

lucasvaltl (approved):

PR #2786, opened 2022-09-23
Remember the last accessed tab in `<Ide-toggle>`

filiptronicek:
## Description
When you change to a tab like <kbd>Jetbrains</kbd> in the IDE toggle, the browser will remember your choice and on the next load, it will default to that choice for tabs. This means you as a user don't have to choose <kbd>Jetbrains</kbd> on every toggle just because you don't use VS Code Browser. It also remembers your changes across the active tab and also any other tabs of docs you may have open.

| Across one tab with refresh | Across multiple tabs |
|--|---|
| remembertab | remembertabs | 

## How to test
Try the behavior on a page like Config ports as described in the description.

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2786"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2786"></a>


src/lib/components/docs/ide-toggle.svelte
```diff
@@ -31,14 +33,53 @@
   ];
   let activeValue = 1;
 
-  const clickHandler = (tabValue: number) => () => (activeValue = tabValue);
+  const rememberLastAccessed = (value: number) => {
+    try {
+      localStorage.setItem("ide-toggle", items[activeValue - 1].slotName);
```
loujaybee, line 38: **nit:** I often think it's better to make a dictionary of keys for localstorage somewhere central, so properties aren't updated randomly across the codebase.

filiptronicek: That's actually really clever, thanks!

filiptronicek: This change made it in https://github.com/gitpod-io/website/pull/2786/commits/41ec7e33e05aae13a007ad94a4a747877387e8b6

PR #2809, opened 2022-09-28
Reuse the old teaser image that got misplaced somehow

axonasif:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
```


Siddhant-K-code (changes requested):
Currently, there is already a image which is getting rendered. I think, we have to delete [this image](https://github.com/gitpod-io/website/blob/main/static/images/blog/i-said-goodbye-to-local-development-and-so-can-you/teaser.png) also.

Can you do that little change also?

axonasif:
Thanks for pointing that out @Siddhant-K-code, I'm actually a bit confused because I didn't notice that file before! But now it's there inside my Gitpod workspace but not on Github when I try to explore from here 😂

Siddhant-K-code (approved):

PR #2825, opened 2022-09-29
2022.09 Self-Hosted Release - Changelog Entry

lucasvaltl:
## Description
Describes the 2022.09 release in terms of features and breaking changes.

## How to test
- Review contents

## Release Notes

```release-note
NONE
```

### DIFF
--- a/src/lib/contents/changelog/2022-08-26 copy.md
+++ b/src/lib/contents/changelog/2022-08-26 copy.md
@@ -1,0 +1,31 @@
+---
+title: September Self-Hosted Release
+excerpt: We've released a new version of Self-Hosted Gitpod 🎉.
+date: 2022-09-30
+image: 2022-09-30.jpg
+tag: self-hosted
+alt:
+---
+
+<script>
+  import Contributors from "$lib/components/changelog/contributors.svelte";
+  import Badge from "$lib/components/changelog/badge.svelte"
+</script>
+
+We've released a new version of Self-Hosted Gitpod. Update instructions can be found in our [update guide](https://www.gitpod.io/docs/configure/self-hosted/latest/updating). You can read more about how to install it in our [documentation](https://www.gitpod.io/docs/configure/self-hosted/latest). More details on this release can be found on [GitHub](https://github.com/gitpod-io/gitpod/releases).
+
+> **Note:** If you are on a paid [Self-Hosted plan](../self-hosted), this release will be promoted to your release channel in one week.
+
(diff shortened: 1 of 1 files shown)


src/lib/contents/changelog/2022-09-30.md
```diff
@@ -0,0 +1,46 @@
...
+
+#### Additional safeguards during updates
+
+[#13215](https://github.com/gitpod-io/gitpod/pull/13215) During updates, we recommend that you do not have any workspaces running to prevent data loss (see [update guide](https://www.gitpod.io/docs/configure/self-hosted/latest/updating)). There is now a preflight check that checks for running workspaces before you deploy and fails if it finds any running ones. If an update is deployed anyways, running workspaces will be stopped for you. Note that users will see their workspaces stopped arbitrarily.
+
+#### We've refactored large parts of the KOTS installation job
+
+[#12196](https://github.com/gitpod-io/gitpod/issues/12196) The KOTS installer script is used to generate a config file out of the values inserted into the KOTS UI. We've refactored large parts of it to turn it from a bash script into a part of the installer image. This improves reliability and predictability for you and maintainability for us. It also allows us to call this during preflight checks, enabling us to provide the preflight checks for config patches mentioned above.
```
lucasvaltl, line 35: How much do we want to talk about internal refactorings like this? Not saying we should not have this, but in theory a user should not have to care about this. On the flip side, those technically interested will find this interested. 🤷

lucasvaltl:
Note that this should be /hold until https://github.com/gitpod-io/gitpod/issues/13494 is sorted

src/lib/contents/changelog/2022-09-30.md
adrienthebo: My inclination is to not keep it, or put it down in the **Fixes and Improvements** sections. I quite like the perspective of sharing details for the technically interested/inclined, but they're likely to read through the entire release notes. Keeping more actionable details earlier in the release means cluster admins are more likely to read through the installation notes that are most relevant to them.

Siddhant-K-code (commented):
Please Change the Filename. There is same [filename](https://github.com/gitpod-io/website/blob/main/src/lib/contents/changelog/2022-09-30.md) for SaaS September Changelog Release

adrienthebo (approved):

src/lib/contents/changelog/2022-10-04.md
```diff
@@ -0,0 +1,45 @@
+---
+title: September Self-Hosted Release
+excerpt: We've released a new version of Self-Hosted Gitpod 🎉.
+date: 2022-10-04
```
adrienthebo, line 4: Flagging this as a future concern, we'll need to update the release date when 2022.09.0 ships properly.

kylos101 (approved):
:ship: it

PR #2869, opened 2022-10-07
feature: add Getting started with nuaware webinar page

jacobparis:
## Description

Adds registration page for Getting Started with Gitpod for Nuaware

/webinars/getting-started-with-nuaware 

image

## Related Issue(s)

Fixes #

## How to test

Register for the event
Check your inbox for a confirmation email

## Release Notes

```release-note
```

## Documentation

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2869"></a>

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2869"></a>


jacobparis:
@metcalfc the page is set up now, and registrants end up in a Google sheet

Just need to update info for the hosts, currently showing Chris

I can do this if you give me the info or you should be able to see where to add that yourself if you'd like

image

chrifro:
Nice ✨ Great to see so many events taking place 🥳

**Some thoughts:**
- is there a need to introduce a new label? Can we not use one of the existing ones e.g. webinar or discussion panel?
- hosted by Gitpod and organized by Gitpod + unaware seems confusing. Let's always mention both?
- we usually use the CET timezone, I'd suggested adding it here as well
- I can sign up multiple times with the same account. We usually restrict that to one sign-up per email address
- I didn't receive a confirmation email. The email flow should be in place before launching the page
- remember to review if the "additional resources" are the ones that bring the most value or if they should be swapped

metcalfc:
@jacobparis Is there anything left to check? Happy to help however I can to ship it. :shipit:

jacobparis:
> @chrifro 
> * is there a need to introduce a new label? Can we not use one of the existing ones e.g. webinar or discussion panel?

This is a distinct type of event to the ones we've had before, so we should add a new label. However if there's interest in keeping our number of labels small, I think "workshop" works well for this and will be reusable in the future

> * we usually use the CET timezone, I'd suggested adding it here as well
> * remember to review if the "additional resources" are the ones that bring the most value or if they should be swapped

In this case, it's a semi-private event for Nuaware folks, all of whom are in the PDT timezone, so it'll be more useful to them to stick to that

> * hosted by Gitpod and organized by Gitpod + unaware seems confusing. Let's always mention both?

Good catch, done

> * I can sign up multiple times with the same account. We usually restrict that to one sign-up per email address

This is a really good catch, but for other reasons. When I dove into how we were restricting them to one signup, it turned out it was only checking for unique timestamps because that was column 0. We were reusing the contact submit form with a condition that checks if it's a webinar and if so does not send the contact email. With the way this was set up beforehand, using the `signup` type as a heuristic for both which column we should check uniqueness of and also as a way of determining if it's a webinar, it would have broken email sending for all general support requests. 

Those concerns have been separated, we can now specifically determine which column should be unique, and there's no chance of it breaking our other forms now. 

> @metcalfc Is there anything left to check? Happy to help however I can to ship it. :shipit:

We're all good now, ready to merge

PR #2880, opened 2022-10-11
fix: color style of multiline codeblocks

Siddhant-K-code:
This PR updates the `selection` css style of multi-line codeblock.

fixes [Internal Discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1665236347843519)

> Signed-off-by: Siddhant Khare <siddhant@gitpod.io>

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2880"></a>


chrifro:
I don't think this is an improvement
|before| after|
|-|-|
|Bildschirmfoto 2022-10-11 um 12 17 11|Bildschirmfoto 2022-10-11 um 12 16 55

Now the highlighted text looks better on light theme, but weirdly on dark theme. 
I'd say we have more urgent issues to address and that we can keep the color highlighting as it is right now. I'd suggest closing this PR.

For next time and in order to be more valuable of our time, let's open an issue first. We'll then discuss and prioritize it during the website sync and reflect the prioritization in the website board. 

When there is time to work on a new issue, it's best to pick an issue from the prioritized column of the [website board](https://github.com/orgs/gitpod-io/projects/28/views/4) cc @Siddhant-K-code 🙌

Siddhant-K-code:
@chrifro why did it gets closed? we can put in draft

chrifro:
See my comment above:
>Now the highlighted text looks better on light theme, but weirdly on dark theme.
I'd say we have more urgent issues to address and that we can keep the color highlighting as it is right now. I'd suggest closing this PR.


I don't see us addressing this in the next weeks and it shouldn't be in the priority column of our upcoming work. If you think we create value by keeping it in our drafts anyway (regardless of upcoming merge conflicts) you can do so.

PR #2906, opened 2022-10-14
fix: github tooltip visibility

### LINKED ISSUE #2903: Bug: GitHub tooltip doesn't show up
issue_gif

When hovering on GitLab and Bitbucket icons, the tooltip shows up. 

## Expected
When hovered, it should show "GitHub logo" as tooltip

Asmit2952:
## Description

GitHub tooltip wasn't visible when we hovered over it. This was because we weren't using direct svg file, instead we were using <GithubMark> tag.

Modified the svg file to work with both light and dark modes. We can use this svg across the site as well.

## Related Issue(s)

Fixes #2903 

Preview video:

https://user-images.githubusercontent.com/81866614/195828539-de373e1c-53b9-449c-9960-ca291b41646e.mp4

Additional notes:
Changed GitHub svg to more accurate one (including colors). The left logo is the new one, and the right logo is the old one. 

| light mode | dark mode |
|----------|:-------------:|
| Screenshot (65) | Screenshot (64) |

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2906"></a>


Cahllagerfeld (commented):
Hey 👋,
thanks for your contribution 🚀

Here's some context around the current implementation:
The icon in the previous version was imported as an SVG, and  not as an image, because of the darkmode. When having the icon imported as an svg we can utilize tailwind classes which makes the handling for light/dark-mode easier.

I added a screenshot from the deployment-preview:
grafik
I think the contrast between icon and background isn't high enough in dark-mode.

I'd love to loop in @chrifro here: 
- We're having two approaches. Either use the solution we had before, and handling the colors on our own with tailwind clases - by doing this we would lose the alt-text again (the reason the issue got created)
- The second would be to have two versions of the icon, and hide/show the correct one based on the state. By following this approach we could also swap to the new icon set mentioned above

As the icon has a fairly small size, I'd be fine with both approaches in that scenario. Before merging we would only need to check where the icon is used elsewhere so we don't break things on other places in the app.

Asmit2952:
> Here's some context around the current implementation: The icon in the previous version was imported as an SVG, and not as an image, because of the darkmode. When having the icon imported as an svg we can utilize tailwind classes which makes the handling for light/dark-mode easier.

The icon was indeed imported as svg, but in the final render the tooltip wasn't visible. What I did was, modified the already present [github.svg](https://github.com/gitpod-io/website/blob/main/static/svg/github.svg) by adding this piece of code in svg file itself for handling light/dark mode.
```
<defs>
    <style>
      svg {
        fill: #333333;
        color-scheme:light dark;
      }
      @media (prefers-color-scheme:dark) {
        svg {
          fill: #f5f5f5;
        }
      }
    </style>
  </defs>
```
 
> I added a screenshot from the deployment-preview: grafik I think the contrast between icon and background isn't high enough in dark-mode.

The [deployment preview](https://deploy-preview-2906--gitpod-kumquat.netlify.app/) looks fine for me?
Screenshot (66)

And thank you for explaining the implementation and current approaches :)

Asmit2952:
Hey @Cahllagerfeld any updates on this PR?

Asmit2952:
Hey @Cahllagerfeld any updates on this PR?

Cahllagerfeld:
Hey @Asmit2952, 
sorry for the long waiting time, I think I missed that notification, unfortunately...

Let me drop another thought I had. 
In addition to the two suggested things above there might be a third one: 
- Turn all Svgs on that page into inline SVG and add a `title` tag to the SVG. By doing this we wouldn't require an alt-text anymore as the SVG gets rendered inline which is more predictable than loading them as an image. I think this solution might do the trick for us, as we keep control over the SVG-colors

Does this make sense to you?

Nonetheless, we should think about upgrading the Github-SVG if there is a new version available.

Re-tagging @chrifro for visibility here.

Siddhant-K-code:
Closing this, in favor of #3324

PR #2943, opened 2022-10-25
quick-fix: remove `beta` label from vim

### LINKED ISSUE #2942: Vim beta label on homepage misleading
### Page URL

https://www.gitpod.io/

### Description

Currently Vim is shown as "beta" on the home page but we don't since we don't have an explicit integration for Vim outside of the SSH CLI integration right now, I would suggest that we remove the label here, as there's no "beta" to speak of (in other words, Vim is not not part of any [product release cycle](https://www.gitpod.io/docs/help/public-roadmap/release-cycle)). 

image

Siddhant-K-code:
fixes #2942

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2943"></a>


loujaybee (approved):

loujaybee:
🙏

PR #2954, opened 2022-10-26
Replace mention of "private key" with "public key"

atduarte:
## Description

As pointed out by a user, we incorrectly mention "private key" where we should mention "public key".

<a href="https://gitpod-staging.com/#https://github.com/gitpod-io/website/pull/2954"></a>


Siddhant-K-code (approved):
Thanks ✨

PR #2969, opened 2022-10-28
nit: SEO fix

Siddhant-K-code:


Siddhant-K-code:
AllY, BP & SEO back to 💯

src/lib/components/footer/index.svelte
```diff
@@ -138,8 +138,8 @@
           <FooterLink href="/cookies">Cookie policy</FooterLink>
         </li>
         <li>
-          <!-- svelte-ignore a11y-missing-attribute -->
           <a
+            href="/"
```
Cahllagerfeld, line 142: I guess that should be a button instead of `<a>` now, whenever clicking that link in the footer you're taken to the index page, although you may not intend to do that by clicking there. I'm not sure if that's the intended behavior 🤔

Siddhant-K-code: The button was creating unnecessary space in the mobile view.

Cahllagerfeld: But then it's a styling issue, isn't it? In my opinion the behavior how it is now isn't the intended one 🤔

Siddhant-K-code: I've created a fix: https://github.com/gitpod-io/website/pull/2970. Can you please take a 👀 & ✅

PR #2986, opened 2022-11-03
Download fix

Cahllagerfeld:


src/routes/blog/future-of-software-cdes.md
```diff
@@ -14,7 +14,7 @@ title: The future of writing software happens in Cloud Development Environments
 
 <script>
   import Quotes from "$lib/components/blog/cde-quotes.svelte"
```
Siddhant-K-code, line 16: above line should also have `;`

Cahllagerfeld: Yeah, I just wonder if this PR is still required, I tested the thing it should fix a couple of minutes ago, and everything worked 🤔

PR #2992, opened 2022-11-04
docs: outputs of required `gp` CLI commands

### LINKED ISSUE #2944: Add gp CLI command output examples to reference documentation
### Description

Add `gp` command output examples to reference page: https://www.gitpod.io/docs/references/gitpod-cli

I was just adding code block examples of CLI output in a recent edition of the changelog, and it got me thinking that it would also be good to have examples in the reference page, also. You can easily see the output of the command.  

image

Siddhant-K-code:
fixes #2944

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/2992"></a>

### DIFF
--- a/gitpod/docs/references/gitpod-cli.md
+++ b/gitpod/docs/references/gitpod-cli.md
@@ -53,6 +53,16 @@ Example to start a interactive guide of `.gitpod.yml` configuration file:
 gp init -i
 ```
 
+**Example output**
+
+```sh
+Use the arrow keys to navigate: ↓ ↑ → ←
+? Workspace Docker image:
+  ▸ default
+    custom image
+    docker file
+```
+
 ## open
 
 Modern editors/IDE's support command line tooling to open a file (e.g. VS Code `code foo.txt`). In Gitpod, this can be done using `gp open <filename>`.
@@ -68,7 +78,7 @@ gp open README.md
 
 `gp preview` opens a URL. The default is to show the URL in a preview pane within the editor or IDE directly. Alternatively, you can show the preview a new tab with the `--external` flag.
 
-Make sure you provide a valid URL, i.e. including the protocol. For example, http://localhost:8080.
+Make sure you provide a valid URL, i.e. including the protocol. For example, <http://localhost:8080>.
 
 You can also use `gp preview <url> --external` to open the URL in a new browser tab.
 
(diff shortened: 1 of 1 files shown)


gitpod/docs/references/gitpod-cli.md
```diff
@@ -202,6 +242,24 @@ Interact with workspace timeout configuration. You can learn more in [Life of a
 gp timeout
 ```
 
+**Example output**:
```
andreafalzetti, line 245: Maybe, we should consider removing the example for `gp timeout` instead of trying fixing the confusion by adding the sub-commands helper.

A similar example is`gp tasks`, we don't have a snippet for `gp tasks` alone but only for actual sub-commands like `gp tasks list`.

This one, makes me think we're suggesting to execute: `gp timeout` which does nothing.
Screenshot 2022-11-09 at 08 24 08

gitpod/docs/references/gitpod-cli.md
```diff
@@ -277,12 +379,24 @@ Here's an example which will make port `3000` public:
 gp ports visibility 3000:public
 ```
 
+**Example output**:
```
andreafalzetti, line 382: Are these examples adding real value to the user?

I like the consistency between all commands, but maybe we should consider where "Example output" really makes a difference? 🤔 

I am suggesting to reflect on this because by looking at the current version, it looks cleaner / easier to go through.


| Current                                                                                                                                                              | After this PR                                                                                                                                                        |
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Screenshot 2022-11-09 at 08 28 58 | Screenshot 2022-11-09 at 08 29 08 |


Alternatively, we could consider having the output as comment in the snippet box?

Somethin like:

```bash
gp ports visibility 3000:public

# Example output:
# port 3000 is now public`
```

loujaybee: Thanks a lot @andreafalzetti 😍  for the great, and challenging thoughts as always, much appreciated. I don't have time today to look and formulate a good reply here, so if you can give me a day or two and I'll get back to this. Feel free to put in draft @Siddhant-K-code if you need in the mean time.

andreafalzetti:
@Siddhant-K-code @loujaybee I didn't mean to block this improvement for long, so if you think it's right, just go ahead and merge 🙏

PR #3002, opened 2022-11-05
Link multiline tasks to error handling

abitrolly:
Added a notice pointing to error handling to multiline tasks and shortened text a little.

An open question is if `one task` == `one terminal`, or if there could be multiple tasks running for the same terminal?

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3002"></a>


Siddhant-K-code:
> if one task == one terminal, or if there could be multiple tasks running for the same terminal?

There could be multiple steps under a task & it can be executed in a single terminal but in the order, they are defined.

Siddhant-K-code (approved):
Thanks ✨

PR #3023, opened 2022-11-14
partial-fix: correct filepath for `Edit in Gitpod`

### LINKED ISSUE #948: Edit in Gitpod - Invalid path for `*/index.svelte` files
Docs pages such as `src/routes/docs/configure/index.svelte` or any other `*/index.svelte` page break the logic in the `src/components/docs/edit-in-gitpod.svelte` component.

A solution, if it gets implemented, is https://github.com/sveltejs/kit/issues/269.

Otherwise, we have to think of a different approach and may be able to leverage meta information provided by Vite :thinking:.

Siddhant-K-code:
## Description

This PR Proposed to partially fixed #948. It would fix all the pages that are there in children. But, this doesn't seem to work for the parent's `index.md` file, because this logic doesn't seem to work => 

https://github.com/gitpod-io/website/blob/b21e0f4a72bd888043d62bfcaef9338d818e4b97/src/lib/components/docs/edit-in-gitpod.svelte#L7-L10

Any possible suggestions @Cahllagerfeld?

## Related Issue(s)

Related #948 

## How to test

- Open Preview
- Go to `/docs`
- Click or Hover on "Edit in Gitpod" button to check the URLs

## Release Notes

```release-note
NONE
```


Cahllagerfeld:
Could you please test if it works? :)

Siddhant-K-code:
It's working 💯

Let's :shipit:

PR #3024, opened 2022-11-14
Events Update

### LINKED ISSUE #3022: follow-up improvements for /events
- [x] reduce skateboard size (see [mockup](https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=7230%3A12039&t=jwT1JFbspGTGBmZx-1))
- [x] add /events to the resources drop-down below "community"
- [x] update event cards: remove headlines as they are already part of the image. Downside: the title will not be crawled by search engines. But I think that's okay, as the description text includes many keywords
- [ ] label upcoming and past webinars. How can we make sure this data stays up-to-date? Do we have a checklist once an event has passed? e.g. swap the registration form with YouTube the embed and update labels. What would be the best way to maintain this? 
- [x] update conference cards: remove headline, add year label, swap primary button style with secondary button style
- [x] update highlights from devx conf cards: swap primary button style with secondary button style

Cahllagerfeld:
closes #3022

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3024"></a>


src/lib/contents/dropdown.ts
```diff
@@ -28,4 +28,9 @@ export const links: {
     href: "/screencasts",
     description: "Short videos to get started",
   },
+  {
+    text: "Events",
+    href: "/events",
+    description: "Latest Events",
```
Cahllagerfeld, line 34: Hey @chrifro, which text do we want to display in the header dropdown?

chrifro: What dropdown do you mean? In case you mean this text, it can stay the same
Bildschirmfoto 2022-11-14 um 15 01 38

Cahllagerfeld: The dropdown under resources :)

chrifro: Ah you mean in the navigation dropdown 💡 


```
**Events**
Upcoming events and event recordings 
```


cc @MichaelAring

Cahllagerfeld:
From my side it's ready to review. The only nit is the gray pills with the dates in dark mode. Should I use `#514f4d` for them?

chrifro (approved):
> From my side it's ready to review. The only nit is the gray pills with the dates in dark mode. Should I use `#514f4d` for them?

Yes, let's color labels for the year in #514F4D for dark theme

One more change: Let's use secondary buttons instead of labels for the event cards. I think that will make the call to action more obvious. Sorry for not deciding on that earlier 🙈
Bildschirmfoto 2022-11-14 um 15 47 23

Other than that, it looks good to me

PR #3034, opened 2022-11-16
Make Card clickable

Cahllagerfeld:


chrifro:
I noticed that the buttons have a weird shadow but only on dark theme 
cc @Cahllagerfeld 
Bildschirmfoto 2022-11-18 um 12 12 16

PR #3041, opened 2022-11-17
feat: Introduce Cypress test on each PR

Siddhant-K-code:
## Description

This PR proposes the Cypress test automation workflow on each PR and removes the tests execution on each workspace start.

## Related Issue(s)

Fixes #[Internal Discussion](https://gitpod.slack.com/archives/C01KGM9ETU6/p1666953603725489)

## How to test

See Below :eyes: in Actions details

## Release Notes

```release-note
None
```

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3041"></a>


.github/workflows/cypress-e2e-test.yml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/checkout@v2
+      - name: Install dependencies 📦
+        uses: cypress-io/github-action@v4
+        with:
+          install-command: npm install
+          runTests: false # just perform install
+      - name: Start Dev Server 🏗️
+        uses: cypress-io/github-action@v4
+        with:
+          start: npm run dev
```
Cahllagerfeld, line 17: When creating a new Sveltekit app with the newer versions Playwright gets delivered as Testing-Framework out-of-the-box. When running the tests, what they do is build the app with `npm run build` and run the tests against the production build. What do you think about this?

Siddhant-K-code: we eventually need to start the server to run tests:

image

Siddhant-K-code: Production Build sever starting issues !!! 

image

Siddhant-K-code:
Cypress doesn't seem to connect to our forwarded port!
I think, we should continue with the dev server only, as it would be a computation building in Action & then also in netlify. WDYT @Cahllagerfeld

Cahllagerfeld:
Why do we need a forwarded port here? 😅 Isn't it like technically running it on localhost? 🤔

Siddhant-K-code:
> Why do we need a forwarded port here? 😅 Isn't it like technically running it on localhost? 🤔

not sure, why it kept failing in cypress 🤷🏼‍♂️

Siddhant-K-code:
@Cahllagerfeld, Tests are passing now, I think there was an issue (resolved in #3080). PR looks good to go :shipit: 

WDYT?

.github/workflows/cypress-e2e-test.yml
```diff
@@ -0,0 +1,24 @@
+name: Cypress End-to-end tests 🧪
+on: [pull_request]
```
Cahllagerfeld, line 2: When is this event fired? Is it only when the PR is opened, or also whenever someone pushes to the PR?

Siddhant-K-code: it will get trigger to both PRs and also new commits to the PRs 😄

.gitpod.yml
```diff
@@ -19,13 +19,8 @@ ports:
 # List the start up tasks. You can start them in parallel in multiple terminals. See https://www.gitpod.io/docs/configure/workspaces/tasks/
 tasks:
   - init: npm install
+  - name: Dev Server
     command: npm run dev
-    name: Dev
-  - command: |
-      gp ports await 3000
-      export CYPRESS_BASE_URL=$(gp url 3000)
```
Cahllagerfeld, line 26: Do you think keeping the CYPRESS_BASE_URL still makes sense, in case we want to run the tests in the workspace?

Cahllagerfeld (commented):
Small comment, other than that it looks good to me 🚢

PR #3077, opened 2022-11-29
Sveltekit Upgrade

### LINKED ISSUE #3075: Upgrade Sveltekit  to `1.0.0-next.340`

Cahllagerfeld:
closes #3075

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3077"></a>


Cahllagerfeld (commented):
Do you think we should update other dependencies as well, when touching them already? @Siddhant-K-code

src/lib/components/startups/organisations-list.svelte
```diff
@@ -29,7 +29,7 @@
 <ul class="space-y-micro">
   {#each organisations as { name, url }}
     <li>
-      <a href={url} target="_blank">{name}</a>
+      <a href={url} target="_blank" rel="noreferrer">{name}</a>
```
Cahllagerfeld, line 32: Also did some minor tweaks at multiple places

Siddhant-K-code:
> Do you think we should update other dependencies as well, when touching them already? @Siddhant-K-code

Yes, a lot of dependencies are outdated.

svelte.config.js
```diff
@@ -53,9 +53,9 @@ const config = {
     },
     trailingSlash: "never",
     adapter: adapterNetlify({
-      split: true,
+      edge: false,
+      split: false,
```
Cahllagerfeld, line 57: We need to turn this off, otherwise, the functions aren't bundled correctly. I'm not sure if that is intended, or if it's a bug in that version.

Siddhant-K-code: I think, it also got resolved in further versions.

Siddhant-K-code:
[`cd5ae23`](https://github.com/gitpod-io/website/pull/3077/commits/cd5ae23b62411c3f9d48c01828f7f9b007bf5bf3) was just a small merge conflict fix, occurred due to #3086

Siddhant-K-code:
It will trigger the Cypress CI also 😄

PR #3091, opened 2022-12-02
Update changelog header image for November 2022

gtsiolis:
## Description

Update changelog header image for November 2022, to include the prebuilds changes.

See [relevant discussion](https://gitpod.slack.com/archives/C022HAT0H5Z/p1669977069496889?thread_ts=1669803102.279449&cid=C022HAT0H5Z) (internal). Cc @jldec


src/lib/contents/changelog/2022-11-30.md
```diff
@@ -2,8 +2,8 @@
 title: November Changelog
 excerpt: November Changelog
 date: 2022-11-30
-image: 2022-11-30-og.jpg
-ogImage: 2022-11-31-og.jpg
+image: 2022-11-30.jpg
+ogImage: 2022-11-30.jpg
```
gtsiolis, line 6: **thought**: Not sure if there's a convention we follow for file names or why we're using a different image for the open graph image. @Cahllagerfeld Do you know?

Cahllagerfeld: The Opengraph image is the one that is used inside this Changelog as the header. We decided to use two different ones, because when sharing it on twitter or discord the image with the bullet list would already give a brief overview of the changes

jldec:
Thanks @gtsiolis

PR #3113, opened 2022-12-07
Update team page

gtsiolis:
## Description

This will update:
1. The **team members count** following the handbook data, see screenshot below and [team members list](https://www.notion.so/gitpod/5d95f8cc71ad440da8d1de27b35c45da?v=6225c1a2a769409fbf65808e89df62e9) (internal).
2. Replace the countries metric with **countries and regions**, because **remote work** is not just about which countries we hire in, but also regions within the same country in which wouldn’t have been possible otherwise.

Screenshot 2022-12-07 at 10 39 20 PM

See also [relevant discussion](https://gitpod.slack.com/archives/C01L357FM9R/p1636111414014800) (internal).

Cc @shannonkinney95 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3113"></a>


src/lib/contents/team.ts
```diff
@@ -44,13 +44,13 @@ export const teamImages: Image[] = [
...
   },
   {
     main: "From",
-    sub: "23",
-    third: "countries",
+    sub: "53",
+    third: "countries and regions",
   },
   {
     main: "Distributed across",
```
gtsiolis, line 56: **question**: How do we get the number of timezones? 🤷

Cc @chrifro because of the [relevant discussion](https://gitpod.slack.com/archives/C03DHEN7VE3/p1663001083391849) (internal).

src/lib/contents/team.ts
```diff
@@ -44,13 +44,13 @@ export const teamImages: Image[] = [
...
     third: "remote team members",
   },
   {
     main: "From",
-    sub: "23",
-    third: "countries",
+    sub: "53",
+    third: "countries and regions",
   },
   {
```
gtsiolis, line 55: **question**: @chrifro What do you think of removing the **time zones** and **continents** elements?

I believe it's causing us too much friction to maintain these numbers, as it's not easy to look them up because the current tools we use can not provide these numbers easily. This is making it challenging to iterate with small MVC changes like this. See [relevant discussion](https://gitpod.slack.com/archives/C03DHEN7VE3/p1670569817509589) (internal).

| BEFORE | AFTER |
|-|-|
| Screenshot 2022-12-09 at 11 16 14 AM (2) | Screenshot 2022-12-09 at 11 16 28 AM (2) |

Cahllagerfeld:
Do we need any updates here, or can we merge @gtsiolis ?

gtsiolis:
@Cahllagerfeld The timezones and continents could be outdated, but I don't think there's an easy way to find this.

I think we've already spent too much energy and time on this. ❗ 

Taking a bias for action here to move forward with a boring (simple) solution by removing the timezones and continents elements 96328f0 as described in https://github.com/gitpod-io/website/pull/3113#discussion_r1044245046 for the reasons included there. 🛹 

Feel free to open a follow-up issue to discuss this further, or restore the removed elements in a follow-up PR. 🏓 

Thanks, everyone! 🏀

Issue #1416, opened 2022-01-03
feature table: add a hover style so it's easier see the different rows


nisarhassan12:
![image](https://user-images.githubusercontent.com/46004116/147940060-c888a59a-abbd-46b6-84d3-dbaa85e9c40f.png)

Cahllagerfeld:
Shall we tackle this with #1384, when we recreate the whole component, or do this before? The concern I'm having is, that the component is currently organized in columns and not rows. Do we need to restructure the whole component for this, to work with rows? @nisarhassan12

mikenikles:
@chrifro Do you think this is something we want to implement?

Siddhant-K-code:
I am not in support of this. 

@chrifro wdyt?

chrifro:
Yes, this is very nice to have especially for longer tables. The Figma component can be found [here](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=2056%3A4219).

Siddhant-K-code:
Reopening this due to #3298

PR #3114, opened 2022-12-08
Adjustments to Endpoint

Cahllagerfeld:


chrifro (commented):
Great work! 🔥

A few very detailed picking comments:

- [x] the width of the pricing card should be slightly smaller (368px instead of 384px). Do they then fit next to each other on tablet view?
- [ ] when selecting "contact sales" there should be a new pre-selected subject for "Enterprise" on /contact/sales (see [internal discussion](https://gitpod.slack.com/archives/C035763SNSE/p1670343662169469) cc @Siddhant-K-code )
- [ ] cost estimator: update messaging to clarify that there is no free plan for teams cc @MichaelAring 

- [ ] cost estimator: should we change the setting for the total members to max +100? or +500? Otherwise, most of the slider returns the same output and it's more difficult to adjust it for smaller quantities cc @MichaelAring 
- [x] cost estimator:  let's add a "," for numbers >999 e.g. `$4,396`
- [ ] the cost estimator should ideally have the same width as the pricing cards (see mockup) -> not a crucial change, but we need to watch out to reuse existing widths
- [ ] previously we distinguished between € and $ based on the user's location. Can we continue doing that? It could be a follow-up task as well
- [ ] it's confusing that we call the Enterprise plan "Gitpod dedicated" in the feature table cc @MichaelAring 
- [x] feature table: tooltips are not readable. The line should break after approx 220px
Bildschirmfoto 2022-12-08 um 09 23 08

- [ ] as a follow up we should work on #1416 to make the table easier to read
- [x] FAQ: update design for "existing customers on old pricing model" (see [mockup](https://www.figma.com/file/K0VV8lc3pmlMVModYM70nj/Website?node-id=8203%3A15236&t=ZXKCgLQazRPDrt3w-1))

**Mobile:**
- [ ] Generally, the spacing doesn't seem to be aligned with the [mobile spacing guidelines ](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=2864%3A2856&t=m0igrMS5Mi2pepMj-1) e.g. it can be reduced between the pricing cards & FAQs, around the "used by "-section, between the cost estimator and its headline, etc.

Bildschirmfoto 2022-12-08 um 09 35 10
 

- [ ] the category headers are too big on mobile. It should be 16px instead of 24px
Bildschirmfoto 2022-12-08 um 09 38 17

- [ ] the spacing inside the cards isn't meaningful (aligned with the hierarchy) and required way too much scrolling. Let me know if you need a mockup to improve on this
Bildschirmfoto 2022-12-08 um 09 42 43

- [ ] after clicking "hide features" the viewer jumps to a totally different part of the website. The viewport should be aligned with the "compare features" section

- [ ] the text field width inside the cards are not aligned. e.g. 
for FAQ it's 360px card width and 312px text field width
for pay-as-you-go card its 360px card width and 266px text width
for the pricing cards its 344px card width and 282px text field width
We need to have a consistent style throughout the website (could also be a follow-up issue)

**Dark theme:**

- [ ] the stroke around the cost estimator is missing (it's likely the same for light theme, but more obvious on dark theme)
Bildschirmfoto 2022-12-08 um 09 54 20

PR #3122, opened 2022-12-08
Nit fixes

Siddhant-K-code:
nit fixes :smile:


chrifro (approved):
Sorry for the back and forth, but let's stick to #1A1712 for the right side of the estimator card. I think the shadow was missing before. It's too black with the new color. Let's stick with #1A1712 and keep the shadow
Bildschirmfoto 2022-12-08 um 15 38 15

Otherwise it LGTM

Cahllagerfeld:
> But, Yeah! I got the point, will change it. Made that change in mine alraedy 🙏

chrifro:
Ah I should have looked at the changed files and not just at the screenshots. Would also be helpful if you outlined in the description that you only made changes to the messaging :) 

`Get started by creating your team. Only pay what you use.` Doesn't make sense in this context and should be removed

Issue #1439, opened 2022-01-07
Document Incremental Prebuilds and Incremental Workspaces


jankeromnes:
### Describe your feature

We should document Incremental Prebuilds 😁

They attempt to make new Prebuilds faster by using a Prebuild from an older commit as a base (as opposed to starting from scratch, which is what happens for non-incremental Prebuilds, or when no suitable parent Prebuild can be found).

When a new Prebuild is started for a Project's commit, and the Incremental Prebuilds feature is enabled for this Project, Gitpod will look for a "compatible" Prebuild from an ancestor of this commit by fetching and "walking back" the history of the commit (Gitpod currently fetches 100 ancestor commits, because anything prior to that is considered obsolete).

"compatible" means a few things (e.g. the parent Prebuild should be already finished and successful; it should have the same `init` / `before` / `prebuild` tasks, it should not be incremental itself, etc.)

The precise parent-selection logic is implemented here:

https://github.com/gitpod-io/gitpod/blob/91c09f7c2bfae33fc6085e48c22f897409a16732/components/server/ee/src/workspace/workspace-factory.ts#L66-L113

Would be cool to translate this into natural language in our docs.

### How to test

1. Go to a Gitpod Project's Settings, and enable Incremental Prebuilds
2. Start a new Prebuild (e.g. by pushing a commit to a branch, or by clicking the `Run Prebuild` button)

jldec:
Prioritizing since we're talking about using incremental prebuilds more
e.g. in https://github.com/gitpod-io/gitpod/issues/13754

PR #3145, opened 2022-12-12
[docs] Document Incremental Prebuilds and Incremental Workspaces

jankeromnes:
## Description

Document Incremental Prebuilds and Incremental Workspaces.

## Related Issue(s)

Fixes https://github.com/gitpod-io/website/issues/1439

## How to test

- https://deploy-preview-3145--gitpod-kumquat.netlify.app/docs/configure/projects/incremental-prebuilds
- https://deploy-preview-3145--gitpod-kumquat.netlify.app/docs/configure/projects/last-successful-prebuild

## Release Notes

```release-note
[docs] Document Incremental Prebuilds and Incremental Workspaces
```

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3145"></a>


jankeromnes:
Screenshot 2022-12-13 at 17 06 58

@Cahllagerfeld @loujaybee do these two new sub-menu entries make sense from a "new docs organization" perspective? 👀 

@gitpod-io/engineering-webapp does the content of the two new pages make sense? 🙂

gitpod/docs/configure/projects/use-last-successful-prebuild.md
```diff
@@ -0,0 +1,40 @@
...
+
+Similar to [incremental prebuilds](/docs/configure/projects/incremental-prebuilds), Use Last Successful Prebuild leverages older prebuilds in order to create new workspaces faster.
+
+![Use Last Successful Prebuild](../../../static/images/docs/projects/use-last-successful-prebuild.png)
+
+When enabled, new workspaces may start based on a prebuild that was built for an older commit, in order to reuse any existing build artifacts. This is achieved by:
+
+1. Loading the older prebuild
+2. Updating the source code to the latest commit
+3. Running any `before`, `prebuild`, and `init` tasks again interactively (incremental rebuild)
```
geropl, line 22: I feel it's worth highlighting here that this means we might run the `init` task twice, so it needs to be re-entrant.

:cloud: We should also update the "lifecycle of a workspace" we have somewhere, and add a asteriks which points to this paragraph... but that's maybe out-of-scope for now. :+1:

gitpod/docs/configure/projects/incremental-prebuilds.md
```diff
@@ -0,0 +1,45 @@
...
+
+Similar to [use last successful prebuild](/docs/configure/projects/use-last-successful-prebuild), incremental prebuilds leverage older prebuilds in order to create new prebuilds faster.
+
+![Enable Incremental Prebuilds](../../../static/images/docs/projects/enable-incremental-prebuilds.png)
+
+An incremental prebuild is based on an older prebuild that was built for an older commit, in order to reuse any existing build artifacts. This is achieved by:
+
+1. Loading the older prebuild
+2. Updating the source code to the latest commit
+3. Running any `before`, `prebuild`, and `init` tasks again (incremental rebuild)
```
geropl, line 20: Same goes for `init` here. :point_down:

geropl:
Thanks a ton @jankeromnes for tackling this - super helpful to be able to link to it. :sparkles: 

Regarding the structure: Ideally I feel both entries should be a section under "Prebuilds". But given the amount of complexity that's in there, and the fact that we want to reference these prominently (at least for now), I think this is good as-is. :+1:

Siddhant-K-code:
Thanks @jankeromnes for writing these ✨ 

> do these two new sub-menu entries make sense from a "new docs organization" perspective? 👀

Agree with Incremental Prebuilds 👍🏽 
for the second one, i think we could use `Last Successful Prebuild` to keep it short make it more ally. See below:

|Before|After|
|---|---|
|image|image|


Or We can simply go with Gero's suggestions by keeping both as a section on `/docs/configure/projects/prebuilds` Page

gitpod/docs/configure/projects/use-last-successful-prebuild.md
jankeromnes: Thanks for the feedback! (To me, this information was already contained in the word "again", but since this is quite subtle and has a big impact, I agree it might be worth highlighting more strongly.)

I'll add this in-line (but it could also be a dedicated call-out, e.g. with an icon and a different background):

> note: this means that your init tasks will be run twice on the same checkout, i.e. they need to be [reentrant](https://en.wikipedia.org/wiki/Reentrancy_(computing))

jankeromnes:
Many thanks for the helpful reviews! 🙏

> Regarding the structure: Ideally I feel both entries should be a section under "Prebuilds". But given the amount of complexity that's in there, and the fact that we want to reference these prominently (at least for now), I think this is good as-is. 👍

Thanks. This was also my thinking (both would make more sense as sections inside the Prebuilds page -- but that one is already so big that I felt they would just add to the complexity / get lost in the noise). Separate sections for now works for me too, and maybe later we'll want to refactor them.

Siddhant-K-code:
@jankeromnes Please use `.webp` images for screenshots 🙏🏽 

 You can following [this guide](https://github.com/gitpod-io/website/blob/main/docs/adr/0003-optimal-assets-format.md) to do so. Let me know, if you need any help.

Siddhant-K-code (approved):
LGTM from the website's point-of-view. 

- Menu items aligned ✅ 
- Optimal Assets ✅

jankeromnes:
Many thanks! I'll go ahead and merge this first iteration so that we have the new docs sections in place.

Happy to keep iterating on them if we think of further improvements to make. 👍

jldec:
Thanks for these super-helpful docs @jankeromnes !

PR #3170, opened 2022-12-16
Fix Currency

Cahllagerfeld:
bring back old logic for displaying the currency

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3170"></a>


Cahllagerfeld (commented):
I brought the old logic we used before back. I tested it with the tailscale Us-Exit node but it keeps displaying in Euro.

I'm not sure if this is/was working as expected before because of that.

Siddhant-K-code (approved):
Looks good from technical perspective ✨ 

/hold for @MichaelAring approval

Then we can :shipit:

Siddhant-K-code:
@Cahllagerfeld We could ship this as we got the approval [here](https://gitpod.slack.com/archives/C027AUU7BC3/p1671203628187479) _(internal)_


cc: @MichaelAring

PR #3184, opened 2022-12-20
Updating gitpod.io/security

securitymirco:
This pull request is adding a reference at gitpod.io/security to our security faq and trust center.

It also fixes some minor details
- referring to Chris as our "Head of Engineering"
- our "view Self-Assessment" button, to be in line with the rest of the page


src/routes/security/index.svelte
```diff
@@ -33,6 +33,17 @@
     <LinkButton variant="primary" size="medium" href="#program"
       >View security program</LinkButton
     >
+    <LinkButton
+      variant="primary"
```
Cahllagerfeld, line 37: All of the buttons are having the styling of a primary cta. Could we potentially re-arrange them, so we don't have 3 primary buttons next to each other? 

cc @chrifro

securitymirco:
@Cahllagerfeld @chrifro do you think we can merge this PR? There is still an open question regarding the buttons being marked as primary

Siddhant-K-code:
> do you think we can merge this PR? There is still an open question regarding the buttons being marked as primary

@securitymirco, I think we should not block the PR. Regarding Button Style we could iterate on later on. Let's :shipit: it 🚀

PR #3198, opened 2022-12-22
chore: bump `postcss` dependencies

Siddhant-K-code:


Cahllagerfeld (commented):
While testing, how would I test that the issue we try to fix with this change, is actually fixed? 😅 The reason I'm asking is that we're bumping major versions for multiple packages, which means we need to be extra careful to not break anything

PR #3199, opened 2022-12-22
Let robots index the `/quickstart` page :wrench:

Siddhant-K-code:


Cahllagerfeld (commented):
I think this is working as expected already 🤔 Also when going to `/quickstart`, there's no `noindex` in the meta-tags of the page

PR #3215, opened 2023-01-05
design-polish: careers page

Siddhant-K-code:
## Description

Since, It's an external link.

|Before|After|
|:---:|:---:|
|image|image|

## How to test

- Open Preview 
- Open /careers


chrifro (commented):
The usual external link style doesn't apply here as it's not an embedded text link but sort of a button itself. It's a bit of an unusual exception. 

The linked page is also a page that is owned by Gitpod and not a classic external page. 

Therefore I'd discard this change and keep things as they are :)

Issue #1950, opened 2022-04-27
use box shadows consistently with the component library


nisarhassan12:
In the component library, we only have 3 box shadows while on the live site we have a total of 5 we should make it so that the 3 mentioned in the component library are used consistently. 

```scss
--shadow-light: 0 1px 1px rgba(0, 0, 0, 0.1);
--shadow: 0 0.625rem 1.5rem 0rem rgba(0, 0, 0, 0.08);
--shadow-medium: 0 1.25rem 2.5rem rgba(0, 0, 0, 0.1);
--shadow-brand: 0 0.625rem 1.5rem var(--brand-quaternary);
--shadow-light-1: 0px 3.58399px 8.95998px rgba(0, 0, 0, 0.08),
    0px 8.95998px 35.8399px rgba(0, 0, 0, 0.12);
```

chrifro:
Some shadows also don't align with the component library, see this example
|live|mockup|
|-|-|
|<img width="600" alt="Bildschirmfoto 2022-06-15 um 08 30 09" src="https://user-images.githubusercontent.com/46345125/173758033-c5427df8-a153-4032-ac14-47b20cbe5ee6.png">|<img width="395" alt="Bildschirmfoto 2022-06-15 um 08 30 18" src="https://user-images.githubusercontent.com/46345125/173758074-8068eafe-e8cd-43d7-9566-c1f81fbe0c1a.png">
|

see https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=514%3A2831

nisarhassan12:
@chrifro Do you mean using only 4 shadows for the light mode and 3 for the dark as per the Figma file? and getting rid of any other shadow being used and switching that to use one of the values from the shadows defined in the mockup ?

chrifro:
Exactly. These are the style that should be in use
<img width="232" alt="Bildschirmfoto 2022-07-05 um 15 46 39" src="https://user-images.githubusercontent.com/46345125/177342799-3146892f-bf2c-4445-b60f-7c9489425a36.png">

If you find any shadows that are not close to any of the specifics in the component library, let me know and we will find a solution :)

chrifro:
I think some shadow styles have already been updated by now. Some have not. E.g. labels in dark theme should follow the button shadow style
[...]

PR #3239, opened 2023-01-12
Add Tags to blog

### LINKED ISSUE #3021: Add tags to each blog post
### Description

Instead of linking the most recent articles in `/blog`, we should add tags to each blog post and display the three most relevant ones. Related to https://github.com/gitpod-io/website/issues/1683

Cahllagerfeld:
closes #3021

For better UX, SSR gets enabled for the `/blog` route. otherwise the filtering would happen `onMount` which looks a bit weird to the user

Added a new prompt to the `npm run generate:blog-post` when creating a new post

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3239"></a>


src/app.d.ts
```diff
@@ -18,6 +18,7 @@ declare global {
       guides: any[];
       securityLogs: any[];
       educationCustomers: any[];
+      filter: string;
```
Siddhant-K-code, line 21: **(early) suggestion:** can we use something as `tag` or `category`? wdyt?

URL Slug would be something like => `/blog?tag=Engineering`

Cahllagerfeld (commented):
Hey @chrifro I added the tags from notion to the corresponding blog posts. There are still a lot of posts missing without one. Do we need to add them within this PR, or shall we do this as a follow up if required?

chrifro:
Hi @Cahllagerfeld , the ones mentioned in the RFC are just some examples. Can you add the tags to all other blogs post? If you are uncertain about any, you could make a table overview for them and we figure it out together. I'd say all blog posts since 2021 should have a tag.

chrifro (commented):
Awesome @Cahllagerfeld  ✨


### **Some comments:**
**on/blog**

- [x]  the hover for the filters is missing

Bildschirm­foto 2023-01-16 um 09 17 17

- [x]  I used the [secondary button style](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=513%3A527&t=tWnkDc3sL5udmUrq-1) which has a defined height of 32 px. Your button is 37px in height. What's your intention behind that?

- [ ] on dark theme: the label shadow is off. Can also be addressed in #1950 [see [component library](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=557%3A1088&t=tWnkDc3sL5udmUrq-1)]

Bildschirm­foto 2023-01-16 um 09 24 38
 
 **On /blog/xyz**:

- [ ] when being on a blog post about "engineering" the "you might also like section" should only display other blog posts with the same label. Let's display 3 random blog posts with the same label from the last 12 months. 
 
 
 - how can we make sure that a new tag will be added for new blog posts? Any ideas? 
 
unrelated: 
 - the minimum spacing between text and authors should be 32 px
Bildschirm­foto 2023-01-16 um 09 30 14

Cahllagerfeld:
I'm a bit on the fence with picking random blogposts to display because the result won't be predictable. We may link three different ones after rebuilding the page, which would lead to inconsistencies also with SEO, because the linked pages now link to other pages.

I don't know if this is a bad impact or not though.
Ideally we would have a way to have 3 different posts per post, but in a predictable way I think 🤔

Cahllagerfeld:
- added a prng for being able to predict the result of the random generate for the blogposts displayed in the section
- updated the button to use the actual button component
- updated spacing for avatars

chrifro (approved):
Works great! Awesome solution for displaying the same random blog posts ✨

Let's merge it 🚀

PR #3253, opened 2023-01-16
feat: check valid company domain

### LINKED ISSUE #3249: Replace company name with company website and add validation to the website input field in contact forms
**Goal**
Add a validation rule that makes sure input in company website fields is an actual domain - doesn't need the www. but should make sure that there is a suffix like `.com` . This should be done for all contact forms where we have the company or website input field. If there is a `company name` or `company` input field (e.g. on https://www.gitpod.io/contact/support)  it should be changed to company website

**Context:**
Domain input is required for enriching leads - if the input is not a valid domain then our chances of success are significantly lower

image

Siddhant-K-code:
## Description

|Invalid|Valid|
|:---:|:---:|
|image|image|

Affects:

- `/contact/support`
- `/contact/sales`
- `/cde`
- `/dedicated`
- `/blog/introducing-gitpod-dedicated`
- `/blog/future-of-software-cdes`

## Related Issue(s)

Fixes #3249

## How to test

- Open [Preview](https://deploy-preview-3253--gitpod-kumquat.netlify.app/)
- Go to below mentioned pages & try to fill invalid domain as shown in image above

- [`/contact/support`](https://deploy-preview-3253--gitpod-kumquat.netlify.app/contact/support)
- [`/contact/sales`](https://deploy-preview-3253--gitpod-kumquat.netlify.app/contact/sales)
- [`/cde`](https://deploy-preview-3253--gitpod-kumquat.netlify.app/cde)
- [`/dedicated`](https://deploy-preview-3253--gitpod-kumquat.netlify.app/dedicated)
- [`/blog/introducing-gitpod-dedicated`](https://deploy-preview-3253--gitpod-kumquat.netlify.app/blog/introducing-gitpod-dedicated)
- [`/blog/future-of-software-cdes`](https://deploy-preview-3253--gitpod-kumquat.netlify.app/blog/future-of-software-cdes)

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3253"></a>


src/lib/components/ui-library/input/input.svelte
```diff
@@ -59,7 +59,13 @@
     {...$$restProps}
   />{/if}
 {#if hasError}
-  <legend class:error={hasError} class="text-xs block mt-1 mb-2">
-    Please fill out the {name ? name : ""} field
-  </legend>
+  {#if id === "company-website" || id === "company"}
+    <legend class:error={hasError} class="text-xs block mt-1 mb-2">
+      Please fill out the {name ? name : ""} field with a valid domain
```
phimae, line 64: Can we just call this "Please enter a valid domain" to make it shorter?

phimae (commented):
Thanks for the speedy execution @Siddhant-K-code! Just left one comment around shortening the error message

Wondering why the preview environment hast different drop downs for Cloud Infrastructure than what is in production?

src/lib/components/dedicated/signup.svelte
```diff
@@ -119,6 +119,15 @@
       console.error(error);
     }
   };
+
+  // if company website contains dot then it's a valid domain else its invalid
+  $: {
+    if (formData.companyWebsite.value.includes(".")) {
```
Cahllagerfeld, line 125: I think the check works, my idea would have been to use the `URL`-Api of the browser to check, see the [docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL). Let me know what you think about this

Siddhant-K-code: But, that also requires `https://` as a required input.

Siddhant-K-code:
> @phimae Wondering why the preview environment has different drop downs for Cloud Infrastructure than what is in production?

On Which Page?

phimae:
> > @phimae Wondering why the preview environment has different drop downs for Cloud Infrastructure than what is in production?
> 
> On Which Page?

/support

Siddhant-K-code:
It's the same, right? 🤔 

image

image

phimae:
> It's the same, right? 🤔
> 
> image
> 
> image

oh you're right. @Cahllagerfeld changed all of them actually but most have forgotten this one. See[ internal thread](https://gitpod.slack.com/archives/C01KGM9ETU6/p1670845136884229?thread_ts=1670497407.482069&cid=C01KGM9ETU6)

Cahllagerfeld:
I'll add the fix to this PR

Siddhant-K-code:
> oh you're right. @Cahllagerfeld changed all of them actually but most have forgotten this one. See[ internal thread](https://gitpod.slack.com/archives/C01KGM9ETU6/p1670845136884229?thread_ts=1670497407.482069&cid=C01KGM9ETU6)

I think, we can merge this PR as it is unrelated to this PR. wdyt @phimae

src/routes/contact/support.svelte
```diff
@@ -19,7 +19,7 @@
   import { scrollToElement } from "$lib/utils/helpers";
   import { page } from "$app/stores";
   import {
-    cloudPlatforms,
+    dedicatedCloudPlatforms,
```
Cahllagerfeld, line 22: Changed the dropdown Values here as well

src/lib/components/dedicated/signup.svelte
filiptronicek: I agree with @Siddhant-K-code for the less-intrusive validation concerning protocols, **but** this means now that `.` is a valid input, and many other things you cannot do in a domain name as well. From what I've used, I would recommend `isFQDN` from `validator` so that we don't have to trouble ourselves with the nuances that come from something like domain validation. 

Because this is not just about domains (we want to support full URLs as well), we could use it in combination with `new URL()` or just use the `isURL` method from the same `validator` library - it has an option to not require a protocol and would help. 

https://github.com/validatorjs/validator.js/blob/master/src/lib/isFQDN.js
https://github.com/validatorjs/validator.js/blob/master/src/lib/isURL.js

Siddhant-K-code: Resolved in [`7481297`](https://github.com/gitpod-io/website/pull/3253/commits/74812971779a92e947c8a152c3a3cb3111ccbbe6). Please review 🙏🏽 

// cc: @filiptronicek, @Cahllagerfeld

PR #3257, opened 2023-01-17
feat: performance improvement w/ lottie

### LINKED ISSUE #2135: move away from lottie animations.
At the moment the edit and share workspace animations are powered by Lottie the problem with this is that the Lottie player Js library is huge in size i.e `334.7 KB` of JavaScript. We use the `lottie-interactivity`(18.5 KB) package to trigger the animations when the element is brought into the viewport.

If the animations are to be done with Vanilla CSS and JavaScript then that would result in orders of less magnitude of code being shipped resulting in making the load time faster.

Siddhant-K-code:
## Description

We saved loading a whole big chunk JS files of Lottie 😄 

## Related Issue(s)

Fixes #2135

## How to test

Open Below Pages & See 👀 

Affected Pages:

- [`/`](https://deploy-preview-3257--gitpod-kumquat.netlify.app/)
- [`/vs/jetbrains-space`](https://deploy-preview-3257--gitpod-kumquat.netlify.app/vs/jetbrains-space)
- [`/for/enterprise`](https://deploy-preview-3257--gitpod-kumquat.netlify.app/for/enterprise)
- [`/for/education`](https://deploy-preview-3257--gitpod-kumquat.netlify.app/for/education)
- [`/for/startups`](https://deploy-preview-3257--gitpod-kumquat.netlify.app/for/startups)

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3257"></a>


Siddhant-K-code:
I've Tried so many workarounds It's doesn't look like a good tradeoff.

image

Siddhant-K-code:
⚠️ While Doing testing. I saw an error in the browser console:
 
image


Also, Currently in prod. Animations are not working up to the mark. Since animations are not even rendering perfectly 😞 (_ignore the first error_)


image

Siddhant-K-code:
## Before

https://user-images.githubusercontent.com/55068936/214111761-cbd2b137-8620-4db3-bcef-53e36680ee0b.mov



## After


https://user-images.githubusercontent.com/55068936/214111635-11ca2b34-c0cf-4afd-9451-fb078a9be11f.mov

Siddhant-K-code:
Research 🧵 

- Currently, there is no good way to use `lottie-interactivity` with svelte. They provide hooks for React, Vue. But not for Svelte! So, We are still loading ~18kb of minified `lottie-interactivity.js` 
- We have a win with PR, that we no longer using heavy `lottie-player.js` (~400kb) 
- Previously, Animations were not working as required. Not Animations will loop when people scroll it to that component.

src/lib/components/lottie.svelte
```diff
@@ -36,6 +38,6 @@
   background="transparent"
   speed="1"
   {style}
-  loop={false}
+  loop={true}
```
Cahllagerfeld, line 41: I think we didn't play it as a loop before, and just run the animation once. Please correct me if I'm wrong here @chrifro

src/lib/components/lottie.svelte
```diff
@@ -1,6 +1,8 @@
 <script lang="ts">
   import type { Lottie } from "$lib/types/lottie";
 
+  // Known Issue with Lottie & Server Side Rendering: https://github.com/airbnb/lottie-web/issues/2851
+  if (typeof window !== "undefined") import("@lottiefiles/lottie-player");
```
Cahllagerfeld, line 5: We can probably use Sveltekit `browser` [[1](https://kit.svelte.dev/docs/modules#$app-environment-browser)] here. I didn't test if its available in our version though, but I think it is.

src/lib/components/lottie.svelte
chrifro: Correct. Let's also keep it this way. Otherwise, the animations will become too distracting

cc @Siddhant-K-code

chrifro:
Is there anything specific I should look out for? 

I looked at the preview and couldn't notice any difference to the production site. Improving the load time doesn't seem to have a negative effect on the animations. So LGTM

PR #3266, opened 2023-01-18
Docs Mobile Menu

Cahllagerfeld:


src/lib/components/main-nav/mobile-menu/mobile-docs-dropdown.svelte
```diff
@@ -7,17 +7,6 @@
...
 
-<style lang="postcss">
-  .nav-items > :global(*:not(:last-child)) {
-    @apply border-b border-divider;
-  }
-
-  /* for last-child of docs items to create a extra space*/
-  .nav-items-docs > :global(*:last-child) {
-    @apply pb-6;
-  }
```
Siddhant-K-code, line 18: I intentionally added that & used it below to add more space b/w "Help Center" & Templates.

Siddhant-K-code: We need extra padding on the last child of `mobileDocsLinks`

Cahllagerfeld: If it's just for that use case, I would add a regular class to the element, and don't use the global decorator for it 🤔, but feel free to adjust, I didn't pay attention to these details to be honest, I was just looking to fix the width of the templates section

Cahllagerfeld: We're having multiple `nav-items > :global()` selectors here. I would personally try to get rid of as much as possible, to make it easier to understand and maintian

Siddhant-K-code: Oh got it now. Can you please add the `pb-6` for the `:last-child`?

Siddhant-K-code (approved):

PR #3288, opened 2023-01-23
docs: Projects API

Siddhant-K-code:
## Description

This PR adds a documentation for Projects API.

- List All Projects
- Get a Project (Only Header, because it is :construction:)
- Create a Project
- Delete a Project

It includes both - HTTP & Go Examples with required paramaters

## How to test

See Preview of `/docs/references/gitpod-public-api#resource-projects`

## Release Notes

```release-note
docs: Projects API in Gitpod Public API _(Alpha)_
```

> Signed-off-by: Siddhant Khare <siddhant@gitpod.io>


gitpod/docs/references/gitpod-public-api.md
```diff
@@ -872,6 +872,262 @@ curl 'https://api.gitpod.io/gitpod.experimental.v1.TeamsService/DeleteTeam' \
...
+{
+  "projects": [
+    {
+      "id": "<PROJECT_ID>",
+      "teamId": "<TEAM_ID>",
+      "name": "<Name of the project>",
+      "slug": "<slug-of-the-project>",
+      "cloneUrl": "<REPOSITORY_GIT_CLONE_URL>",
+      "creationTime": "2023-01-20T18:12:45.442Z",
+      "settings": {}
```
easyCZ, line 933: Given we don't document the settings property, should we remove it from the examples for now?

gitpod/docs/references/gitpod-public-api.md
```diff
@@ -872,6 +872,262 @@ curl 'https://api.gitpod.io/gitpod.experimental.v1.TeamsService/DeleteTeam' \
...
+	}
+
+	response, err := gitpod.Projects.ListProjects(context.Background(), connect.NewRequest(&v1.ListProjectsRequest{
+		TeamId: "<TEAM_ID>",
+	}))
+
+  if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to list projects %v", err)
+		return
+	}
```
easyCZ, line 913: mis-aligned with the above blocked

easyCZ (approved):
LGTM % nits

/hold if you want to address the comments

gitpod/docs/references/gitpod-public-api.md
```diff
@@ -872,6 +872,262 @@ curl 'https://api.gitpod.io/gitpod.experimental.v1.TeamsService/DeleteTeam' \
...
+|   Parameter    |                              Description                               |  Type  |
+| :------------: | :--------------------------------------------------------------------: | :----: |
+|      `id`      |                               Proejct Id                               | string |
+|    `teamId`    |                              Workspace Id                              | string |
+|     `name`     |                          Name of the project                           | string |
+|     `slug`     | Slug of the project (contains no space & separated by <code>-</code> ) | string |
+|   `cloneUrl`   |    Repository Clone Url (Git Context Url - GitHub/GitLab/Bitbucket)    | string |
+| `creationTime` |                 Timestamp when the project was created                 | string |
+
+### Get a project (Coming soon)
```
loujaybee, line 982: **Thought:** I'm starting to wonder whether we should apply `release status` https://www.gitpod.io/docs/help/public-roadmap/release-cycle cycles to endpoints specifically, rather than the full API, to better define what's alpha, public beta, private beta, etc. It might get too confusing, though.

Siddhant-K-code: Since the whole page is under "Alpha". I don't think, it would make much difference until we tag each spec. with label. 
image

gitpod/docs/references/gitpod-public-api.md
```diff
@@ -872,6 +872,259 @@ curl 'https://api.gitpod.io/gitpod.experimental.v1.TeamsService/DeleteTeam' \
...
+
+<APIExample id="api-example">
+
+<div slot="curlExample">
+
+```bash title="cURL"
+curl 'https://api.gitpod.io/gitpod.experimental.v1.ProjectsService/ListProjects' \
+  -H 'content-type: application/json' \
+  -H 'Authorization: Bearer <YOUR_ACCESS_TOKEN>' \
+  --data '{"teamId":"<TEAM_ID>"}'
```
burningion, line 889: You have <TEAM_ID> here, but no example of the type expected. (A string? UUID? Integer?) Let people know what the type is, and maybe even where to find it. (Via a link if necessary).

gitpod/docs/references/gitpod-public-api.md
```diff
@@ -872,6 +872,259 @@ curl 'https://api.gitpod.io/gitpod.experimental.v1.TeamsService/DeleteTeam' \
...
+</div>
+</APIExample>
+
+<br>
+
+**Request Parameters**:
+
+| Parameter | Description |  Type  | Required |
+| :-------: | :---------: | :----: | :------: |
+| `teamId`  |   Team Id   | String |   true   |
```
burningion, line 964: Ah here I see it's a string. What kind of a string is it? Can we give more detail?

gitpod/docs/references/gitpod-public-api.md
```diff
@@ -872,6 +872,259 @@ curl 'https://api.gitpod.io/gitpod.experimental.v1.TeamsService/DeleteTeam' \
...
+
+<br>
+
+**Response Parameters**: It returns the array of projects for the requested team & total number of projects. Each project has the following parameters:
+
+|   Parameter    |                              Description                               |  Type  |
+| :------------: | :--------------------------------------------------------------------: | :----: |
+|      `id`      |                               Proejct Id                               | string |
```
burningion, line 972: misspelling here of project

burningion (changes requested):

gitpod/docs/references/gitpod-public-api.md
loujaybee: > Maybe even where to find it. (Via a link if necessary).

Like the idea of linking to where this can be found! Did want to note that these ID's are not (at least to my knowledge) surfaced anywhere in the product (see issue below). I've added the Public API label to the issue, and we should aim to get this fixed + so we can update the docs before the API is publically accessible beta! 

- https://github.com/gitpod-io/gitpod/issues/15654

gitpod/docs/references/gitpod-public-api.md
loujaybee: Yeah, it's certainly not-blocking for his PR, more a thought passing through my mind as I see these "coming soon" labels. Let's pick up separately / another time. Until we have a candidate for a public beta, it's not particularly relevant, but is something that I'm thinking about.

gitpod/docs/references/gitpod-public-api.md
Siddhant-K-code: > (A string? UUID? Integer?) Let people know what the type is

I followed OpenAPI Spec standards while defining these (because that's our [future goal](https://github.com/gitpod-io/gitpod-public-api-roadmap)). As [they Said here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#data-types:~:text=Formats%20such%20as,is%20not%20specified.). UUID is a `format` & Not a `type`. 

The thing we could do is add a separate table for `format` **OR** just `string (uuid)`


wdyt @burningion @loujaybee

easyCZ: FWIW, any ID we reference should not state what type it is - UUID, or otherwise. The ID should be opaque to the client and we do not expect the client to do any parsing or understanding of the IDs in any way. 
This is pretty standard in APIs as it allows you to change how you represent the IDs without breaking clients in a major way. That's also the reason it's represented as a String, rather than a more specific datatype.

burningion (approved):

Siddhant-K-code:
@loujaybee If you got some time this week, can you please review this? So we can :shipit: 🙏🏽

Siddhant-K-code:
Closing this for now, As we are not actively working on this.

PR #3294, opened 2023-01-24
[blog] building for the long run

svenefftinge:
Sharing the letter from Johannes as a blog post

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3294"></a>

PR #3295, opened 2023-01-24
minor fix for optimal asset

Siddhant-K-code:
followup for #3294 

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3295"></a>


chrifro:
Nice catch! Thanks @Siddhant-K-code 

While the blog post should have a tag, can we ignore it in the logic of suggesting articles in the  “you might also like” section?

(non -urgent & not part of this PR) Any ideas on how we can make the new flow of adding tags to new blog posts more obvious?

Siddhant-K-code:
Tags are already added ✅ 

See [Preview](https://deploy-preview-3295--gitpod-kumquat.netlify.app/blog/building-for-the-long-run) 👀

chrifro:
Yes I saw that. I just want to make sure that the blog post will not be shown in this section
Bildschirm­foto 2023-01-24 um 17 37 15

Siddhant-K-code:
@chrifro Do you think, we should also hide "You might also like" section from this blog?


image

chrifro:
I think it's okay to keep it. Let's merge it. 

Thanks for fixing this right away 🙏

PR #3374, opened 2023-02-03
Update countries and regions count

gtsiolis:
## Description

Minor count change, see PR description in https://github.com/gitpod-io/website/pull/3113#issue-1482835056.

> Replace the countries metric with countries and regions, because remote work is not just about which countries we hire in, but also regions within the same country in which wouldn’t have been possible otherwise.

If we have a concrete and easier way that everyone can grab the number of countries, let's keep the existing count but rephrase the headline to countries for accuracy.

| Screenshot from the single source of truth (SSOT), [Gitpodder list](https://www.notion.so/gitpod/5d95f8cc71ad440da8d1de27b35c45da?v=6225c1a2a769409fbf65808e89df62e9) (internal) |
|-|
| Screenshot 2023-02-03 at 10 53 32 AM |

Cc @Siddhant-K-code because https://github.com/gitpod-io/website/pull/3304.


src/lib/contents/team.ts
```diff
@@ -49,7 +49,7 @@ export const featureTitles: FeatureTitle[] = [
   },
   {
     main: "From",
-    sub: "14",
```
gtsiolis, line 52: **suggestion**: If we have a concrete and easier way that everyone can grab the number of countries, let's keep the existing count but rephrase the headline to countries for accuracy, and close this PR.

Siddhant-K-code, line 52: These numbers were provided by Eva ([here - internal conversation](https://gitpod.slack.com/archives/C01KGM9ETU6/p1674695210832159?thread_ts=1674674634.575009&cid=C01KGM9ETU6))

Siddhant-K-code (approved):

PR #3361, opened 2023-02-02
📄 docs: start workspace with options modal

Siddhant-K-code:
## Description

Documentation for "start a Gitpod workspace with options" modal

## Related Issue(s)

Fixes #3358

## How to test

Open Preview:

- [Getting Started](https://deploy-preview-3361--gitpod-kumquat.netlify.app/docs/introduction/getting-started#start-your-first-workspace)
- [IDEs and Editors](https://deploy-preview-3361--gitpod-kumquat.netlify.app/docs/references/ides-and-editors#start-a-workspace-with-an-ideeditor)
- [VS Code](https://deploy-preview-3361--gitpod-kumquat.netlify.app/docs/references/ides-and-editors/vscode#open-vs-code-desktop-from-the-gitpod-dashboard)
- [Workspace classes](https://deploy-preview-3361--gitpod-kumquat.netlify.app/docs/configure/workspaces/workspace-classes#select-a-workspace-class-for-your-workspace)

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3361"></a>

PR #3377, opened 2023-02-03
docs: new workspace via options w/ url

Siddhant-K-code:
> :warning: Placemen for the current docs is not decided yet! This is for communication (w/ Mike, Andre, Lou) purpose to improve it further.

Related https://github.com/gitpod-io/gitpod/pull/15567 , #3361


loujaybee:
This might fit nicely inside/alongside: https://www.gitpod.io/docs/integrations

Since this functionality helps to create a sort of integration with Gitpod. 

A similar Google page for inspiration: https://cloud.google.com/shell/docs/open-in-cloud-shell

We should potentially look at taking inspiration from: https://gitpodify.com/ (https://github.com/jacobparis/gitpodify)

loujaybee:
Thanks for adding @Siddhant-K-code ! Would be good if we can review and update the naming before we publish.

Siddhant-K-code:
@loujaybee, Putting it in draft for now. 🙏🏼

Issue #3408, opened 2023-02-09
Minor Improvements on Homepage


Siddhant-K-code:
### Description

> followup #3382, #3387

Tasks List:

- [x] the hover and active states of the IDE selector don't align with mockup yet - https://github.com/gitpod-io/website/pull/3426
    - active: bg-> white, stroke 3px and grey, no shadow
    - hover: bg -> white, linear stroke 1px, no shadow
    
    See mockup
<img width="593" alt="Bildschirm­foto 2023-02-10 um 09 24 48" src="https://user-images.githubusercontent.com/46345125/218040902-6e8799c5-dd95-4d47-b6f1-dc8d67f1a410.png">

- [x] the animated quotes should start a bit earlier now that we don't have the contact pop-up anymore - https://github.com/gitpod-io/website/pull/3409

- [x] align icon size. The JetBrains IDE icons are a lot bigger than the other icons. Let's align them to the smaller style. Because of that the spacing inside the cards is also not even - #3428

<img width="918" alt="Bildschirm­foto 2023-02-10 um 09 26 01" src="https://user-images.githubusercontent.com/46345125/218041067-0a0fb4b9-e1c9-4c40-88a9-6591f5828f02.png">

- [ ] responsive layout bug
<img width="1156" alt="Bildschirm­foto 2023-02-10 um 09 26 40" src="https://user-images.githubusercontent.com/46345125/218041176-aa7c0097-bb57-4d17-adb9-9fb0ec330406.png">

- [x] dark theme: change design to this - https://github.com/gitpod-io/website/pull/3436
<img width="637" alt="Bildschirm­foto 2023-02-08 um 15 00 16" src="https://user-images.githubusercontent.com/46345125/217550970-5a8f45b7-92f8-4986-a7e4-99bef2e309fa.png">

- [x] the Vim screenshot needs to be updated (see [here](https://www.figma.com/file/VL0CJpHqcDJVJ4BEkoyRFV/Gitpod-Website-Component-Library?node-id=2056%3A3271&t=Tun0p4nGsulxrvSI-1)) - https://github.com/gitpod-io/website/pull/3424
- [x] the screenshots a slightly moving around when hovering over the different IDEs. We need to align their size and position - https://github.com/gitpod-io/website/pull/3424
- [x] remove white border from screenshots - https://github.com/gitpod-io/website/pull/3424
[...]

PR #3409, opened 2023-02-09
:zap: start logo animation fast

Siddhant-K-code:
Related #3408


src/lib/components/animated-logos.svelte
```diff
@@ -28,7 +28,7 @@
         logos[0].classList.add("hide", "to-top");
         logos[1].classList.remove("hide", "to-top", "to-bottom");
         logos[2].classList.add("hide", "to-bottom");
-      }, 8000);
+      }, 3800);
```
chrifro, line 31: @Siddhant-K-code did you change the speed of the logo rotation as well? It feels way too fast now. Could you revert that? 🙏

Siddhant-K-code: Since We need to start fast and it should be aligned with the start speed for all logos. That's why we need to make it fast. Would find some workaround.

Siddhant-K-code: raised #3422

chrifro: Before the logos wouldn't start before x seconds that's because we had the context widget to pop-up first. Now that we removed the contact pop-up we don't need that big of a time delay any more the start the animation. 

The new start-up time (`await sleep: 900`) works great. But the animation itself is too fast. Please revert it to its original speed

Siddhant-K-code: I have published a change. Looks better now, let me know your thoughts.

PR #3462, opened 2023-02-23
Deprecate Security Log

securitymirco:
This PR removes our old security log at https://www.gitpod.io/security/log in favour of our Trust Center Notifications at https://app.safebase.io/portal/71ccd717-aa2d-4a1e-942e-c768d37e9e0c/preview?product=default


Siddhant-K-code (commented):
We need to remove the page calls from `/changelog`.


as pointed out in the build error

image

PR #3467, opened 2023-02-23
[docs] Update TailScale guide (no need for `USER root` + no need to install `jq` a 3rd time)

jankeromnes:
## Description

- There is no need to switch to `USER root` thanks to `sudo` (which is even already used in the snippet)
- There is no need to tell people to install `jq` because it's already in `gitpod/workspace-full` AND it's even already re-installed a 2nd time in the `apt-get` snippet

## Related Issue(s)

Fixes #

## Release Notes

```release-note

```

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3467"></a>


jankeromnes:
Giving up.

PR #3527, opened 2023-03-20
Filter in product (VS Code) changelog by a custom front-matter property (distinguish ad-hoc vs roll-up

akosyakov:
## Description

Filter in product (VS Code) changelog by a custom front-matter property (distinguish ad-hoc vs roll-up

## Related Issue(s)

fix https://github.com/gitpod-io/gitpod/issues/16897

## How to test

Check /changelog/latest endpoint, play with tag properties from sources to see what it returns. There is some caching, you have to do force reload.

## Release Notes

```release-note
NONE
```


Siddhant-K-code (approved):
Implementation Looks good to me ❤️. Any specific reasons to name it as "rollup"?, Instead of something more intuitive like `major` (this would show on IDEs) & `minor` (this won't show on IDEs)

src/routes/changelog/raw-markdown.ts
```diff
@@ -18,7 +18,7 @@ export const get: import("@sveltejs/kit").RequestHandler = async ({
   const fileName = releaseId
     ? `${releaseId}.md`
     : (locals.changelogEntries as ChangelogEntryType[]).find((e) => {
-        return !e.tag || e.tag.indexOf("self-hosted") == -1;
```
loujaybee, line 21: **Question:** So we didn't have any tagged changelog items as `self-hosted`?  🤔

loujaybee: Does that mean...

* The self-hosted changelog filter was not working before? 
* This part of the code is not respected / not the only place where we filtered? 
* Or possibly, we filtered in the VS Code side?

loujaybee (approved):

src/routes/changelog/raw-markdown.ts
akosyakov: We were tagging, it was working. I inverted a guard here, changelog either self-hosted or rollup now.

PR #3537, opened 2023-03-22
Update Gitpod Workspace image to Jammy

kylos101:
## Description

Update the base image to Jammy. @axonasif , you fixed prebuilds! cc: @jeanp413 

image

image

## Related Issue(s)

Fixes # n/a

## How to test

Start a Gitpod Workspace from this PR, you'll see the tasks actually run now.

## Release Notes

```release-note
NONE
```

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3537"></a>


Siddhant-K-code (approved):
Looks good to me 🙃

axonasif (approved):
It seems to be working good! I just noticed a few changes for vnc on Jammy:
- `/tmp/.X0-lock` is no longer being created by the updated display server, but the vnc startup script checks for it's existence to determine if the vnc server is already running[^1]. Because of that, the startup script is executed for every new terminal even if vnc is running. Also a race condition may exist now when you have many supervisor tasks since the IDE process doesn't start with the login shell anymore for performance reasons (not related with Jammy).
- Upon opening Google chrome, it prompts for a keyring password, not sure if this was the case in Focal, some packages probably changed their config.

These are non blocking for our website repo I think, just thought I'd mention 🙏 

[^1]: https://github.com/gitpod-io/workspace-images/blob/5530267b5da00c49be537ddfd8740b99daeaf217/chunks/tool-vnc/gp-vncsession#L16

kylos101:
@axonasif agreed, please do us a favor, and create those as issues in the workspace-images repo?

PR #3549, opened 2023-03-28
re-route for changelog rss

Siddhant-K-code:
> Thanks @gtsiolis for feedback 🧡

This would redirect `/changelog/rss` to `/changelog/rss.xml`


filiptronicek (approved):
Nice!

PR #3561, opened 2023-04-07
Clarify how workspace dir works

KristerV:


gitpod/docs/configure/projects/prebuilds.md
```diff
@@ -107,6 +107,8 @@ Prebuilds save **only** the workspace directory. Other directories like the home
 
 To use global installs like 'npm install -g', we suggest using a custom Dockerfile or a `before` task, which runs before the init task, and runs again before the command task.
 
+Anything you do in the Dockerfile will be saved across the image, but the `/workspace` directory is actually mounted into the image after the build. So changes you make in a session will only be saved if they're in the `/workspace` directory, but setting up your environment is not limited to this space, in fact it will be overwritten after the image build.
+
```
svenefftinge, line 111: > but setting up your environment is not limited to this space, in fact it will be overwritten after the image build.

I don't understand that last part.

KristerV: m, i just wanted to repeat the fact that you shouldn't put anything in /workspace through Dockerfile, beacuse it's overwritten.

kylos101 (commented):
> the docs in regards to what is persistent is very foggy at best. quite frustrating experience.

This is fair feedback, thank you. Also, my apologies for the supremely late reply!

> what is a "prebuild" anyway? is it the Dockerfile + init? is it just init?

Prebuilds and their relation to the `/workspace` directory are shared [here](https://www.gitpod.io/docs/configure/projects/prebuilds#workspace-directory-only) for posterity.

> m, i just wanted to repeat the fact that you shouldn't put anything in /workspace through Dockerfile, beacuse it's overwritten.

What do you think of us adding a couple docs changes?
1. Under [Learn Gitpod](https://www.gitpod.io/docs/introduction/learn-gitpod), add a section about `Where to save files` (note: for prebuilds, that's already covered [here](https://www.gitpod.io/docs/configure/projects/prebuilds)).
2. Warning users under [Workspace Image](https://www.gitpod.io/docs/configure/workspaces/workspace-image), to avoid making changes to `/workspace` from their dockerfiles.

I can tee them up in a PR, but first wanted to make this makes sense. Let us know?

PR #3600, opened 2023-04-21
:sparkles: Revamp `/enterprise` page

Siddhant-K-code:
fixes #3583


chrifro (commented):
- [x] The alignment of the header image is off. It should not be cut as in the previous image
Bildschirm­foto 2023-04-21 um 14 49 42

- [x] this section needs to be optimized for dark theme

Bildschirm­foto 2023-04-21 um 14 50 30

PR #3617, opened 2023-05-03
Remove individual plan from pricing page

atduarte:
## Description

Update pricing page following org-only mode and the removal of personal accounts.

| Before | After |
|--------|-------|
| Old - Plans and Pricing - Gitpod | Plans and Pricing - Gitpod |

Note that now Enterprise is highlighted at the bottom table in order to align with the highlight at the top table.

## Related Issue(s)

Fixes WEB-295.

## How to test

Check [preview](https://deploy-preview-3617--gitpod-kumquat.netlify.app/pricing).

## Release Notes

```release-note
NONE
```


Siddhant-K-code (approved):
LGTM. 

Hold for @taliamoyal's review

chrifro (approved):
Awesome changes, thanks @atduarte 🥳

non-blocking thoughts:
- what do you think about having the same three columns for the feature table that we use in the top cards? It could be confusing that before only the paid plan was under organizations and in the feature table both the "free" and "paid" plan live under the org
- the main CTA  `Start with 50 hours free` could be misunderstood as everyone has a total of 50h free instead of having them each month. WDYT? Any idea how to rephrase that? :)

chrifro:
> Makes sense! I tried to align it as much as possible but struggled with what to put in the titles. What would you suggest? Both cards with "Organization"?

Why not use the same logic (starter, organization and enterprise)?
Bildschirmfoto 2023-05-04 um 12 21 18


>That I haven't changed, but happy to do so now! Maybe we could simplify with "Start for free"? Or "Start with 50h month free"?

I know, just wanted to put it out there to see if anyone feels the same. If you agree, let's change it to "Start with 50h/month for free"

src/lib/contents/pricing.ts
```diff
@@ -13,44 +13,40 @@ import githubMarkSvelte from "$lib/components/svgs/github-mark.svelte";
...
     btnHref: "https://gitpod.io/login",
-    trackingName: "individual",
+    trackingName: "organization",
   },
   {
     title: "Organization",
     description: "For orgs of any size",
-    showCurrency: true,
-    price: `0.36/hour`,
+    price: "Paid",
```
chrifro, line 30: Idea: what about changing this to "pay-as-go-you"?

chrifro:
Awesome! One more nitpick: the CTAs in the table don't align with the CTAs of the top cards anymore. But we can also follow up on smaller things like that afterward. Your call. Excited to see this live. Thank you 🙌

atduarte:
@chrifro Done 🙌 

@taliamoyal may we proceed? Any changes?

chrifro:
Talia already approved it [here](https://gitpod.slack.com/archives/C01KGM9ETU6/p1683180478194389?thread_ts=1683169103.704889&cid=C01KGM9ETU6) in Slack (internal link)

PR #3635, opened 2023-05-12
Add clarification to Step 3: Gitpodify a project

phimae:
Made the description of step 3 clearer - what is the gitpod.yml and why do you need it


Siddhant-K-code (approved):
Thanks @phimae ✨

PR #3513, opened 2023-03-12
POC: A/B Testing Using ConfigCat Feature Flags

Siddhant-K-code:
> **Warning**: This is a Proof of Concept - POC => No Guarantees for getting it merged & implemented as it is.

fixes #1767

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3513"></a>

PR #3654, opened 2023-05-18
feat: Feature flags for A/B Testing

### LINKED ISSUE #1767: Implement an A/B testing approach
Solutions to investigate:

## GrowthBook
URL: https://www.growthbook.io/
Pros:
- Open source
- Also supports feature flags, which we could use for more than just A/B testing

Cons:
- TBD

Siddhant-K-code:
## Description

Related #3513

## Related Issue(s)

fixes #1767

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3654"></a>


src/lib/utils/feature-flag-provider.ts
```diff
@@ -0,0 +1,6 @@
+import * as configcat from "configcat-js-ssr";
+const configCatClient = configcat.getClient(
+  "OSzbCCd5YESN3ZtcqYIpkA/n8MSGCfKLUC2OcnKX0i2ZQ"
```
selfcontained, line 3: Could we move this to an env file?

Siddhant-K-code: I've tried, but it wasn't seems to be working.

`process.env.CONFIGCAT_KEY` wasn't working 🫠

src/lib/components/index/get-started/index.svelte
```diff
@@ -3,28 +3,37 @@
   import LaunchExampleWorkspace from "./launch-example-workspace.svelte";
   import LinkGitRepository from "./link-git-repository.svelte";
   import Card from "$lib/components/ui-library/card";
+  import configCatClient from "$lib/utils/feature-flag-provider";
 
-  let checked: boolean = false;
+  var featureFlag = true;
```
selfcontained, line 8: There's a mismatch of the default value w/ this approach, where it's `true` here, then in `getValueAsync` it's default is `false`. I think this site is all statically pre-rendered, so not sure this matters too much in practice.

src/lib/utils/feature-flag-provider.ts
selfcontained, line 1: We could improve the ergonomics of getting flags a little by exporting a function something like this:

```ts
import * as configcat from "configcat-js-ssr";

const configCatClient = configcat.getClient(
  "OSzbCCd5YESN3ZtcqYIpkA/n8MSGCfKLUC2OcnKX0i2ZQ"
);

type GetFeatureFlagDone = {
  (flagValue: boolean): void;
};
export const getFeatureFlag = (
  flagName,
  defaultValue: boolean,
  done: GetFeatureFlagDone
): boolean => {
  configCatClient
    .getValueAsync(flagName, defaultValue)
    .then(done)
    .catch((error) => {
      console.error(error);
    });

  // return the default value until the feature flag value is fetched
  return defaultValue;
};
```
Then in the component that needs the flag:

```ts
// Then in the svelte component that needs the feature flag:
import { getFeatureFlag } from "$lib/utils/feature-flag-provider";

let exampleFlagValue = getFeatureFlag(
  "home_example_launch_workspaces_card",
  false,
  (val) => (exampleFlagValue = val)
);
```
This streamlines it a bit, and provides an interface for a single default value.

src/lib/components/index/get-started/index.svelte
```diff
@@ -3,28 +3,37 @@
...
+    });
 </script>
 
 <Section id="get-started" class="pt-20">
   <h2 class="!mb-x-small text-center text-h2">Get started for free</h2>
   <div
     class="cards-container flex flex-col lg:flex-row max-w-[600px] lg:max-w-none gap-4 mx-auto"
-    class:justify-center={checked}
-    class:justify-between={!checked}
+    class:justify-between={featureFlag}
```
selfcontained, line 20: Not related to this PR, but this is cool how svelte supports conditional classes like this. TIL

selfcontained (approved):
Nice job, left some suggestions (probably worth pulling the SDK key out into an env var, and perhaps resetting the key in config-cat since this one in the PR is now in git history. 

With the static rendering that we do here, I suppose the intermittent state of a flag's default value, then the real one being returned isn't really present in the output html/js/css we create during build. Does that sound like the right understanding?

easyCZ:
Whole it's called the SDK Key, it's not actually a secret. It's similar to Google analytics key, which is a public key to target the config.

Given the config is a static file, and eval happens clientside, the worst anyone can do with it is fetch our config file.

Siddhant-K-code:
> the worst anyone can do with it is fetch our config file - @easyCZ 

And, the Code is already open source. So, all the feature flags are already public.

src/lib/components/index/get-started/index.svelte
Siddhant-K-code: Svelte ftw 😉

src/lib/utils/feature-flag-provider.ts
easyCZ: Is there a non-production/preview env for the website? It may make sense to separate the env used between these two, in which case shifting to env vars which are embedded at compile time would make sense. Else, we probs need to switch on domain name or similar.

The value from seperate envs is that you can enable a feature for all preview envs, without risking the same happens in production.

src/lib/utils/feature-flag-provider.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// getFeatureFlag function retrieves the value of a feature flag from the ConfigCat API.
+
+export const getFeatureFlag = (
+  flagName: string,
+  defaultValue: boolean,
+  done: GetFeatureFlagDone
+): boolean => {
+  configCatClient
+    .getValueAsync(flagName, defaultValue, new configcat.User(userId))
```
easyCZ, line 21: If you have other information, such as the region where the user is, it can be useful to add these to the User object as well.

src/lib/utils/feature-flag-provider.ts
```diff
@@ -0,0 +1,29 @@
+import * as configcat from "configcat-js-ssr";
+import { getOrSetCookieId } from "$lib/components/segment.svelte";
+
+const configCatClient = configcat.getClient(
```
easyCZ, line 4: We also configure the following, which are relevant here:
```
 pollIntervalSeconds: 3 * 60, // 3 minutes
        requestTimeoutMs: 2000,
        logger: configcat.createConsoleLogger(LogLevel.Error),
        maxInitWaitTimeSeconds: 0,
```

The `maxInitWaitTimeSeconds` is quite important, as it ensures that the first render is not blocked by fetching feature flags. Unsure exactly how this plays with SSR but worth seeing what the impact is

src/lib/utils/feature-flag-provider.ts
Siddhant-K-code: We just have a Timezone check for EU/US for pricing, here:

https://github.com/gitpod-io/website/blob/af922c41a0f3f6e2d2694e1b8d16722e6063d799/src/lib/utils/helpers.ts#L4-L7


Idk, how would it affect the scenario if we pass on this^ info.

easyCZ: I think ideally you'd want to get this data from either the locale of the browser, or from the LoadBalancer which receives the request based on the user IP. It's not intended to be perfect, but can give you some tools to handle cases where you may want to adjust what you show based on regions. You wouldn't use it to decide things like `$ vs EUR` identifiers, but instead use it for segmentation that you rollout some wording change in EU first, to see if there's a difference between US.

src/lib/utils/feature-flag-provider.ts
Siddhant-K-code: For better way to achieve this, We can use [LazyLoad Polling mode](https://configcat.com/docs/sdk-reference/js-ssr/#lazy-loading) instead of Auto Polling as:

```ts
configcat.PollingMode.LazyLoad,
  {
    cacheTimeToLiveSeconds: 3 * 60
  }
```

wdyt? 💭

easyCZ: Yep, that seems sensible

src/lib/utils/feature-flag-provider.ts
Siddhant-K-code: In netlify, we can provide custom environment variables for Prod. and Preview Envs. But, we never used it separately. All things are aligned (since, it's a static site, not many things get changed/ to test)

Siddhant-K-code: Done in [`5c9ffe4`](https://github.com/gitpod-io/website/pull/3654/commits/5c9ffe4272a19e0bc48800d7c63dfbdd5c078161)

easyCZ (approved):

PR #3659, opened 2023-05-22
:sparkles: design pollish

Siddhant-K-code:
## Description

Design update for https://www.gitpod.io/docs/introduction top card.

Change: CTA Buttons will get left aligned on all screens.

|Before|After|
|:----:|:---:|
|image|image|


chrifro (approved):
Great catch! ✨

PR #3707, opened 2023-06-01
🐛 fix for staging env

Siddhant-K-code:


Siddhant-K-code:
@jakobhero Do you think we should also include `*.netlify.app`? as it would always fail the segment event on staging with `400` status.

https://github.com/gitpod-io/website/blob/920032ceb532c51e48cd27ebd97c41076e2aa24e/src/routes/api/collect-data.ts#L129-L134

src/lib/components/banners/cookie-consent.svelte
```diff
@@ -34,7 +34,6 @@
           on:click={() => {
             Cookies.set(cookies.ANALYTICAL, "false", {
               expires: 365,
-              domain: ".gitpod.io",
```
Siddhant-K-code, line 37: I've removed this explicit check. So, it would verify the cookie where it get created. i.e

If the cookie is generated from `.gitpod.io`. It would get verified on `.gitpod.io`
& vice versa for `.netlify.app` ([Docs](https://github.com/js-cookie/js-cookie/tree/latest#domain))

Siddhant-K-code:
For some reason, it doesn't send any even when a website gets loaded. But, it does send events when clicking on anywhere (CTAs, Navbar, Footer, etc.) 🤷🏼‍♂️

src/routes/api/collect-data.ts
```diff
@@ -126,12 +126,6 @@ export const post: RequestHandler = async ({ request }) => {
             body: { message: "Please include url and path in props" },
             status: 400,
           };
-        if (!body.props.url.startsWith("https://www.gitpod.io")) {
-          return {
-            body: { message: "Invalid URL provided" },
-            status: 400,
-          };
-        }
```
Siddhant-K-code, line 134: Removed it

Siddhant-K-code:
#3712 would also fix the issue :)

PR #3713, opened 2023-06-02
Configure Vite to respect `process.env`

Siddhant-K-code:
Related MKWEB-2

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3713"></a>


vite.config.js
```diff
@@ -39,6 +39,9 @@ const config = {
   preview: {
     port: 3000,
   },
+  define: {
+    "process.env.SEGMENT_KEY": JSON.stringify(process.env.SEGMENT_KEY),
```
Siddhant-K-code, line 43: This is a Hotfix specially for `SEGMENT_KEY`. Ideally, it should be for every Environment variable we are using, for example [this](https://github.com/gitpod-io/website/blob/e0e48dbc942d3b42e0a32c75fc6ed991ab086b5f/src/routes/api/submit-form.ts#L9)

PR #3747, opened 2023-06-15
Not Preload Homepage Image assets & Passive listeners while scroll

Siddhant-K-code:
Fixes MKWEB-23

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3747"></a>


vite.config.js
```diff
@@ -26,7 +27,7 @@ try {
  * @type {import("vite").UserConfig}
  */
 const config = {
-  plugins: [sveltekit()],
+  plugins: [sveltekit(), splitVendorChunkPlugin()],
```
ghostdevv, line 30: Is separating out vendor chunks really a performance benefit?

Siddhant-K-code: Yeah, Not sure. Was just trying to cut up on some on Build time also.

ghostdevv: I'm not sure there is a benefit since we need to load the js anyway

PR #3756, opened 2023-06-16
cross-link `gp validate` docs

Siddhant-K-code:
[IDE-171](https://linear.app/gitpod/issue/IDE-171/link-gp-validate-description-on-gitpod-cli-page)

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3756"></a>

PR #3758, opened 2023-06-16
fix: update docs

loujaybee:
Small update to add description to `gp validate` directly on the CLI page.


gitpod/docs/references/gitpod-cli.md
```diff
@@ -62,7 +62,7 @@ gp init -i
 
 > The `gp validate` command is currently in [Beta](/docs/help/public-roadmap/release-cycle) · [Send feedback](https://github.com/gitpod-io/gitpod/issues/7671).
 
-Learn more about [gp validate](/docs/configure/workspaces#validate-your-gitpod-configuration) in the docs.
+Starts a workspace within in your currently opened workspace to quickly apply and review your configuration changes. Allows you to troubleshoot your `.gitpod.yml`, workspace image, ports, tasks and more. See [configure workspaces](/docs/configure/workspaces) for more on validating a workspace configuration.
```
loujaybee, line 65: We should keep the description + the link, so when users are scrolling through they're not forced to jump to another page 🙏  ([related change](https://github.com/gitpod-io/website/pull/3756/files)). 

Small pedantic one, but if the link text is the title of the other page, a user can make a better judgement about whether the page is interesting to them. Having "configure workspaces" gives more context about what the other page content might be about than "read more" which forces you to click, without knowing if the command is relevant or useful to you. 🙏

Siddhant-K-code (approved):

gitpod/docs/references/gitpod-cli.md
Siddhant-K-code: We should point to specific URL `/docs/configure/workspaces#validate-your-gitpod-configuration` instead of just `/docs/configure/workspaces`

PR #3796, opened 2023-06-28
Align styling on `/support` page

Siddhant-K-code:
fixes [MKWEB-66 : Align `/support` page styles](https://linear.app/gitpod/issue/MKWEB-66/align-support-page-styles)


Siddhant-K-code:
**Note**: for the headline on mobile, we need to do it for `h3` only because anchor link (if we go with `h2`, it would break the mobile layout. e.g.,

image

On `h3` (mobile), it looks better and more readable - 

image

PR #3977, opened 2023-09-19
(docs) add info to decode OIDC token

Siddhant-K-code:
Link to [**preview**](https://deploy-preview-3977--gitpod-kumquat.netlify.app/docs/configure/workspaces/oidc#openid-connect-oidc-in-gitpod)

---

|Before|After|
|:---:|:---:|
|image|image|

<a href="https://gitpod.io/#https://github.com/gitpod-io/website/pull/3977"></a>


src/content/docs/configure/workspaces/oidc.md
```diff
@@ -56,7 +56,7 @@ $ gp idp token --audience example.org
 eyJhbGciOiJSUzI1NiIsImtpZCI6ImlkLTE2ODQ3NTc4MDY...
 ```
 
-The output, when decoded, translates to a JSON object similar to the one below.
+The output, when decoded (using any JWT decoder), translates to a JSON object similar to the one below.
```
loujaybee, line 59: **Suggestion:** Can we give an example here? This feels like it's not so helpful without an example? The user will read this, and then have to go to google to figure this out?

Siddhant-K-code:
> **@akosyakov**: not sure what was motivation but looks ok

Currently. `gp idp token` returns a JWT token which is not human readable. And, there is an extra step the user needs to perform to decode to make it usable information. For future, we can have a optional flag (e.g., `--decode`) in CLI to directly decode and return the JSON output. And, currently, there was no instruction regarding how to decode.

src/content/docs/configure/workspaces/oidc.md
Siddhant-K-code: Resolved in [`f07c7dc`](https://github.com/gitpod-io/website/pull/3977/commits/f07c7dc2d5cdc2f6c4c16ed0385f7a2128f0bd84)

loujaybee (approved):

loujaybee:
Thanks @Siddhant-K-code, a theoretically small tweak, but could make a difference to our developer experience—thank you !

